PR #33, opened 2021-05-21
Update ponca.mdoc

ani37:
a small correction in line 49


nmellado:
Thanks for the PR, but the big cap stand for P-O-N-C-A ;), so it has been done in purpose. 
One thing we could do however is to replace `ponca` by `PONCA` at the same line.

PR #49, opened 2021-08-10
added surface fitting

ani37:


nmellado:
Thanks for the feature.
When evaluated in CI, the `surface_fit` test seems to fail all the time. Is it working on your side ?

PR #55, opened 2022-04-14
Support Range & leaf_count

Aclardgm:
- Add a counter of leaf in kdTree.h, updated during build operations


nmellado (changes requested):
Thanks for your proposal.
Please replace tabulations by spaces.

nmellado (approved):

PR #56, opened 2022-04-20
Factorize code

nmellado:
This PR introduces a massive refactoring of the Fitting library, with the following improvements:
 - In previous versions, the neighbors weight were computed multiple times, this has been fixed. From our experiment, this can lead to an 30% speed-up on the computations.
 - Same when computing Scale/Space/ScaleSpace derivatives: the weight derivatives are now computed once for all. This change required to significantly change the API of the Derivative classes, e.g., 
```cpp
// Plane fitting
using TestPlane = Basket<Point, WeightFunc, CovariancePlaneFit>; 
// Derive Plane in Space
using PlaneScaleDiff = BasketDiff<TestPlane, internal::FitSpaceDer, CovariancePlaneDer>; 
// Both objects have the same API and can now be used with compute methods
```
 - A lot of code was duplicated in the different fit classes. All code is now factorized. A result of this factorization is that intermediate computations (e.g., barycenter) now have their dedicated classes and can be called directly. We still provide helper classes to generate standard fits. The type `CovariancePlaneFit` used above is in fact define as follows:
```cpp
template < class DataPoint, class _WFunctor, typename T>
using CovariancePlaneFit =
CovariancePlaneFitImpl<DataPoint, _WFunctor, // Compute the Plane from the covariance matrix
        CovarianceFitBase<DataPoint, _WFunctor, // Compute covariance matrix
                MeanPosition<DataPoint, _WFunctor, // Compute Mean position (required for the covariance matrix)
                        Plane<DataPoint, _WFunctor,T>>>> // Holds plane-related methods;
```
 - Following the same spirit, we solved several issues limiting the genericity of the library. Primitives and Fitting techniques can now be merged in the same `Basket`, allowing to compute different properties during a single run, without duplicating computation. This feature is designed for advanced users who need custom computation types. The syntaxe is a bit verbose (users need to indicate the intermediate computation objets), but it is the cost to pay to avoid duplicating computation: 
```cpp
using Hybrid = Basket<Point, WeightFunc, 
   AlgebraicSphere, Plane, // primitives
   MeanNormal,             // compute normal averaging: used by OrientedSphereFitImpl
   MeanPosition,           // compute positions averaging: used by OrientedSphereFitImpl and CovarianceFitBase
   OrientedSphereFitImpl,  // Sphere fitting
   CovarianceFitBase,      // compute covariance matrix of the positions: used by CovariancePlaneFitImpl
   CovariancePlaneFitImpl> // Plane fitting
```
 - A side effect of these changes is that we solved several issues with the requirement system, now any wrong combination of computation components should trigger an error.
 - It is now possible to build custom basket, however this also allows to build ill-formed combinations where two computations attempt to save their results at the same place, e.g. when using two fitting technique for the same primitive. This case is not detected at compile time, but rather dynamically when calling `finalize()`, which can now return `CONFLICT_ERROR_FOUND`. Internally, this is implemented by checking if the primitive is already valid (ie. it has been computed already) when finalizing the computation.

### Todo
- [x] Changelog
- [x] Fix issues reported by [Aclardgm](https://github.com/Aclardgm)
- [x] Documentation

### To be added as issues for next release
- [ ] Try to add automatic differentiation type computation


Aclardgm:
Some templates classes in eignet have incoherent behavior using Ponca::GLSGeomvar but not calling geomvar function : 
MLSOperatorTraits
RIMLSOperatorTraits 
and their template instances.

PR #58, opened 2022-04-26
Clear Query Stack 

Aclardgm:
Allow inherited query to clear the query stack and use it more easely


nmellado:
Can you explain why reset is not sufficient ?

nmellado:
@Aclardgm is this still required ?

Aclardgm:
No this is not needed anymore, we can close this request.

PR #70, opened 2023-03-07
[fitting] Adding Wendland and Singular weight kernels

Carl-Delrieu:
Added Wendland and Singular weight kernels to weightKernel.h

### DIFF
--- a/.github/workflows/deployDoc.yml
+++ b/.github/workflows/deployDoc.yml
@@ -26,7 +26,7 @@ jobs:
       - name: Configure Doc
         run: |
           cd build/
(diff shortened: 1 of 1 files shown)


nmellado:
Thanks for the contribution, please find some comments below:

- [ ] Add link to the paper (in each class introduction), mention equation number for clarity.
To this end, add the bibtex entry (you can find it on the ACM website) here: https://github.com/poncateam/ponca/blob/1793b3b2e8b2b59ab30e4e3a3437ecb4e9a448af/doc/src/ponca.bib, and cite as in: https://github.com/poncateam/ponca/blob/1793b3b2e8b2b59ab30e4e3a3437ecb4e9a448af/Ponca/src/Fitting/gls.h#L19
- [ ] Use these functions in the test to be sure everything compile fine, and fitting still work.

NAPondu:
Also fixed a typo in weightKernel.h which was failing GitHub checks

nmellado:
I added the tests in the test suite. 
It passes if I disable tests of the added classes.
Please fix.

nmellado:
Almost there, just need to fix the tests on windows.

PR #72, opened 2023-03-22
Fix barycenter derivatives computation + update doc

nmellado:


nmellado:
@ThibaultLejemble could you please have a look at these changes ?
I am surprised to find such kind of change deep in the library.

Ponca/src/Fitting/mean.h
```diff
@@ -92,12 +94,35 @@ namespace Ponca {
...
+        ///
+        /// We rewrite \f$ t(\mathbf{x}) = \sum u(\mathbf{x})v(\mathbf{x}) \f$, with \f$ u(\mathbf{x}) = w(\mathbf{x}) \f$ and \f$ v(\mathbf{x}) = \mathbf{x} \f$.
+        ///
+        /// By definition, \f$ t'(\mathbf{x}) = \sum u(\mathbf{x})v'(\mathbf{x}) + u'(\mathbf{x})v(\mathbf{x}) \f$, with  \f$ u'(\mathbf{x}) = w'(\mathbf{x})\f$ and \f$ v'(\mathbf{x}) = \mathbf{1}\f$.
+        ///
+        /// Which leads to \f$ b'(\mathbf{x}) = \frac{\sum w(\mathbf{x})\left(\mathbf{1} \sum w(\mathbf{x}) + w'(\mathbf{x})\mathbf{x}\right) - \left(\sum w(\mathbf{x})\mathbf{x}\right)\sum w'(\mathbf{x})}{\left(\sum w(\mathbf{x})\right)^2} \f$
+        ///
+        /// Simplifying by \f$ \sum w(\mathbf{x}) \f$ we obtain:
+        ///
+        /// \f$ b'(\mathbf{x}) = \frac{\mathbf{1} \sum w(\mathbf{x})  + w'(\mathbf{x})\mathbf{x} - b(\mathbf{x})\sum w'(\mathbf{x})}{\sum w(\mathbf{x})} \f$
```
ThibaultLejemble, line 117: I think there is still an error because the derivative of a point is null, so `v' = 0` instead of `v' = 1` (the points denoted here by `x` are actually constant points `x_i` from the point cloud, and not variable with respect to scale nor space).
So `t' = \sum u' v = \sum w' x`, and then `b' = ( \sum w' x - b \sum w' ) / \sum w`. 

I believe this can be compactly computed in one line by `VectorArray barycenterDer = ( m_dSumP - Base::barycenter() * Base::m_dSumW ) / Base::m_sumW`.

nmellado: Thanks for the review, I updated the code and the documentation accordingly, emphasizing the difference between x and p_i.

PR #85, opened 2023-06-05
Use variadic template for basket extensions

azaleostu:
This PR changes the hard-coded extension list for the `Basket` and `BasketDiff` types to a variadic list.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -8,6 +8,9 @@ Current head (v.1.2 RC)
 - Examples
     - Fix example cuda/ponca_ssgls (#109)
 
+- Bug-fixes and code improvements
+    - [fitting] Use variadic template for basket extensions (#85)
+
 --------------------------------------------------------------------------------
 v.1.1 RC
 This release improves the Spatial Partitioning Package with new features and
@@ -31,7 +34,6 @@ improved doc, as well as several bug fixes and API improvements.
     - [fitting] Fix a bug in dtau computation for scale-only derivation (#98)
     - [fitting] Fix a bug in MlsSphereFitDer weight derivatives (#97)
     - [ci] Fix Compiler is out of heap space error on Github (MSVC), by splitting tests (#87)
-    - [fitting] Use variadic template for basket extensions (#85)
 
 - Tests
     - Fix WeightKernel test (failing on windows due to finite differences) (#91)
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -31,6 +31,7 @@ improved doc, as well as several bug fixes and API improvements.
     - [fitting] Fix a bug in dtau computation for scale-only derivation (#98)
     - [fitting] Fix a bug in MlsSphereFitDer weight derivatives (#97)
     - [ci] Fix Compiler is out of heap space error on Github (MSVC), by splitting tests (#87)
+    - [fitting] Use variadic template for basket extensions (#85)
 
 - Tests
     - Fix WeightKernel test (failing on windows due to finite differences) (#91)
--- a/Ponca/src/Fitting/basket.h
+++ b/Ponca/src/Fitting/basket.h
@@ -15,14 +15,60 @@
 namespace Ponca
 {
 
+#define BSKW typename BasketType::WeightFunction
+#define BSKP typename BasketType::DataPoint
+
 #ifndef PARSED_WITH_DOXYGEN
 /*! \brief Namespace used for structure or classes used internally by the lib */
 namespace internal
 {
+    template <class P, class W,
+        typename Aggregate,
+        template <class, class, typename> class Ext,
+        template <class, class, typename> class... Exts>
+    struct BasketAggregateImpl
+    {
+        using type = typename BasketAggregateImpl<P, W, Ext<P, W, Aggregate>, Exts...>::type;
+    };
+
+    template <class P, class W,
+        typename Aggregate,
+        template <class, class, typename> class Ext>
+    struct BasketAggregateImpl<P, W, Aggregate, Ext>
+    {
+        using type = Ext<P, W, Aggregate>;
+    };
+
     /*! \brief Internal class used to build the Basket structure */
-    template <class, class, typename T> class Forward: public T {};
+    template <class P, class W,
+        template <class, class, typename> class... Exts>
+    struct BasketAggregate : BasketAggregateImpl<P, W, PrimitiveBase<P, W>, Exts...>
+    {
+    };
+
+    template <typename BasketType, int Type,
+        typename Aggregate,
+        template <class, class, int, typename> class Ext,
+        template <class, class, int, typename> class... Exts>
+    struct BasketDiffAggregateImpl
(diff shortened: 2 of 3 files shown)


nmellado:
@ThibaultLejemble could you check if this works with CUDA ?

ThibaultLejemble:
> @ThibaultLejemble could you check if this works with CUDA ?

I compiled and ran the Cuda example with this new basket and it works fine! 
There was a compilation issue but not related to this PR (see #109).

Ponca/src/Fitting/basket.h
```diff
@@ -15,14 +15,80 @@
...
+    struct BasketAggregateImpl<P, W, Aggregate, Ext>
+    {
+        using type = Ext<P, W, Aggregate>;
+    };
+
     /*! \brief Internal class used to build the Basket structure */
-    template <class, class, typename T> class Forward: public T {};
+    template <class P, class W,
+        template <class, class, typename> class... Exts>
+    struct BasketAggregate : BasketAggregateImpl<P, W, PrimitiveBase<P, W>, Exts...>
```
ThibaultLejemble, line 45: What is the purpose of this intermediate struct since the class `Basket` could just inherits directly from `internal::BasketAggregateImpl<P, W, PrimitiveBase<P, W>, Exts...>::type`?

Ponca/src/Fitting/basket.h
```diff
@@ -104,40 +165,17 @@ namespace internal
...
-         \tparam Ext10 Implements \ref concepts_computObjectBasketDiff "ComputationalDerivativesConcept" (optional)
-         \tparam Ext11 Implements \ref concepts_computObjectBasketDiff "ComputationalDerivativesConcept" (optional)
+         \tparam Exts Implements \ref concepts_computObjectBasketDiff "ComputationalDerivativesConcept" (optional)
      */
-    template < typename BasketType, int Type,
-            BASKETDIFF_TP(0), BASKETDIFF_TP(1), BASKETDIFF_TP(2), BASKETDIFF_TP(3), BASKETDIFF_TP(4), BASKETDIFF_TP(5), BASKETDIFF_TP(6), BASKETDIFF_TP(7), BASKETDIFF_TP(8), BASKETDIFF_TP(9), BASKETDIFF_TP(10), BASKETDIFF_TP(11) >
-    class BasketDiff
-        : public Ext11<BSKP,BSKW,Type, Ext10<BSKP,BSKW,Type, Ext9<BSKP,BSKW,Type, Ext8<BSKP,BSKW,Type, Ext7<BSKP,BSKW,Type, Ext6<BSKP,BSKW,Type, Ext5<BSKP,BSKW,Type, Ext4<BSKP,BSKW,Type, Ext3<BSKP,BSKW,Type, Ext2<BSKP,BSKW,Type, Ext1<BSKP,BSKW,Type,Ext0<BSKP,BSKW, Type, PrimitiveDer<BSKP,BSKW,Type, BasketType > > > > > > > > > > > > > {
+    template <typename BasketType, int Type,
+        template <class, class, int, typename> class Ext0,
```
ThibaultLejemble, line 171: I think specifying `Ext0` is not necessary

Ponca/src/Fitting/basket.h
```diff
@@ -76,11 +142,6 @@ namespace internal
     }                                                                                                 \
     WRITE_BASKET_SINGLE_HOST_FUNCTIONS
 
-#define BASKETDIFF_TP(I) template <class, class, int, typename> class Ext##I = internal::ForwardDiff
-#define BSKW typename BasketType::WeightFunction
-#define BSKP typename BasketType::DataPoint
-
-
     /*!
          \brief Aggregator class used to declare specialized structures with derivatives computations, using CRTP
```
ThibaultLejemble, line 146: I believe that CRTP is not really used: no class derives from another one using itself as a template argument like `struct Derived : Base<Derived> {};`, right?

Ponca/src/Fitting/basket.h
```diff
@@ -237,7 +261,12 @@ namespace internal
         }
     }; // class Basket
 
-#undef BASKET_TP
+    template <class P, class W, int Type,
+        template <class, class, typename> class Ext0,
+        template <class, class, typename> class... Exts>
+    class BasketAutoDiff : public internal::BasketAutoDiffAggregate<P, W, Type, Ext0, Exts...>::type
```
ThibaultLejemble, line 267: Is the `AutoDiff` feature in draft or can we review it?
Maybe we can move the `AutoDiff` feature in another PR?

Ponca/src/Fitting/basket.h
azaleostu: CRTP isn't used directly for the basket type but it is used to combine all of the extension types into a single aggregate type, with `BasketAggregate`, `BasketDiffAggregate`, etc. (e.g. in `BasketAggregateImpl`, the `Aggregate` template parameter gets passed down to each extension type recursively).

Ponca/src/Fitting/basket.h
azaleostu: I chose to hide the `PrimitiveBase<P, W>` part from the `Basket` class definition as I thought it could be considered an implementation detail, but it is true that what you mentioned is equivalent (I also thought it made the `Basket` class a bit more readable)

Ponca/src/Fitting/basket.h
azaleostu: The class docs mention that `Ext0` is not an optional extension. If it was, `Ext0` would not be necessary, but with just the `Exts` parameter pack the user can provide 0 extensions, which could lead to some nasty template errors with the aggregate types.

Ponca/src/Fitting/basket.h
azaleostu: This was added as a followup to #106. The aggregate part should be ready for review, but I don't know what the functionality of the class is going to be yet, so I can remove this from this PR for now, if necessary.

nmellado: @azaleostu could you remove the autodiff version so we can merge this refactoring, and then add autodiff in another PR (with appropriate testing)

nmellado (changes requested):
Thanks for the proposal.
Please

- [ ] move autodiff out of the propel
- [ ] rebase on `master`
- [ ] update changelog (I can do it if you prefer)

CHANGELOG
```diff
@@ -31,6 +31,7 @@ improved doc, as well as several bug fixes and API improvements.
     - [fitting] Fix a bug in dtau computation for scale-only derivation (#98)
     - [fitting] Fix a bug in MlsSphereFitDer weight derivatives (#97)
     - [ci] Fix Compiler is out of heap space error on Github (MSVC), by splitting tests (#87)
+    - [fitting] Use variadic template for basket extensions (#85)
```
nmellado, line 34: Please move this to line 10

nmellado (changes requested):
Almost there, thx

PR #89, opened 2023-06-13
Update nbNeighbors when using multiple pass

jmaleo:
This little modification fix this bug : 
- When using several pass, the basket multiplies the number of neighbors with the number of pass. 
Be careful, when you're using addNeighbor by hand, don't forget to update the m_nbNeighbors by hand too when using other pass.

### DIFF
--- a/Ponca/src/Fitting/basket.h
+++ b/Ponca/src/Fitting/basket.h
@@ -155,8 +155,6 @@ namespace internal
 
         if (wres.first > Scalar(0.)) {
             Base::addLocalNeighbor(wres.first, wres.second, _nei, dw);
-            Base::m_sumW += (wres.first);
-            ++(Base::m_nbNeighbors);
             return true;
         }
         return false;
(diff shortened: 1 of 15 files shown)


Ponca/src/Fitting/basket.h
```diff
@@ -220,6 +222,7 @@ namespace internal
         WRITE_BASKET_FUNCTIONS;
 
         /// \brief Add a neighbor to perform the fit
+        /// When using by hand, don't forget to set m_nbNeighbors to 0 when performing a new pass (see PONCA::NEED_OTHER_PASS).
```
nmellado, line 225:
```suggestion
        /// When called directly, don't forget to set m_nbNeighbors to 0 when performing a new pass (see PONCA::NEED_OTHER_PASS).
        /// \see compute Prefer when using a range of Points
        /// \see computeWithIds Prefer when using a range of ids
```

nmellado:
I slightly updated your proposal: when reviewing your code, I discovered that we had the same issue with `PrimitiveBase::m_sumW`. I took the opportunity to better handle multi-pass fitting and to hide `PrimitiveBase` internal fields.

PR #93, opened 2023-06-15
Fix - add missing Weight ponderation in covarience fit

AntoineLafon:


nmellado (dismissed):

ThibaultLejemble:
Should we enable compilation warnings to detect such issues ? 
Since `w` was not used we could have seen this in a warning message.
But maybe there is too many warnings and it is difficult to fix them all ?

PR #94, opened 2023-06-16
Change curvature convention to kmin <= kmax

ThibaultLejemble:
This PR changes the convention used to store principal curvatures in the `CurvatureEstimatorBase` class.
Before, `k1` and `k2` where stored such that `|k1| > |k2|`, which was the sources of issues such as discontinuities appearing in visualization.
Now, principal curvatures are renamed in `kmin` and `kmax` and verify `kmin < kmax`.

**Questions** 
Is it the responsability of the caller of `CurvatureEstimatorBase::setCurvatureValues` to ensure that `kmin <= kmax` ? 
Or should the class always check and swap the curvature values and direction if necessary ? 
Right now the first solution is implemented since `Eigen` often compute eigenvalues in this "natural" order, but that can be changed. 

Note that encapsulation is already private. There is still one unused class that try to access them directly using `Base::m_kmin = ...` (`ProjectedNormalCovarianceCurvatureEstimator`), but I think this class will be subject to some cleaning soon (I will investigate what `ProjectedNormalCovarianceCurvatureEstimator` and `NormalCovarianceCurvatureEstimator` are doing exactly).

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -10,6 +10,7 @@ improved doc.
 - API
     - [spatialPartitioning] Add kd-tree traits type (#80)
     - [fitting] Add Primitive::getNumNeighbors (#86)
+    - [fitting] Change naming convention and rational for principal curvatures in CurvatureEstimatorBase (#94)
 
 - Bug-fixes and code improvements
     - [spatialPartitioning] Fix unwanted function hiding with DryFit::setWeightFunc (#86)
--- a/Ponca/src/Fitting/curvature.h
+++ b/Ponca/src/Fitting/curvature.h
@@ -7,7 +7,6 @@
 #pragma once
 
 #include "./defines.h"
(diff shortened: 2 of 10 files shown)


nmellado:
Thanks for the proposal.

> Is it the responsability of the caller of `CurvatureEstimatorBase::setCurvatureValues` to ensure that kmin <= kmax ?

No, I think this should be ensured in the implementation of `setCurvatureValues`, so the convention is an invariant of the class (that an be described in the documentation and ensured whatever happens).

ThibaultLejemble:
I enforced `kmin <= kmax` in the setter. 
I also removed the debug check on the orthogonality of the vector because I am not sure about the right epsilon to use.

nmellado (dismissed):

nmellado (approved):

Issue #95, opened 2023-06-16
MlsSphereFitDer second order weight derivatives are calling convertToLocalBasis twice


ThibaultLejemble:
There is an issue with the second order derivatives of the weight function here :
https://github.com/poncateam/ponca/blob/93a01257e4b5fa320574db4232a225009918d674/Ponca/src/Fitting/mlsSphereFitDer.hpp#L41

They are computed in the local basis in a method `addLocalNeighbor`, but the call to `DistWeightFunc::spaced2w`, `DistWeightFunc::scaledw` and `DistWeightFunc::scaled2w` transform the point in the local basis again. So `convertToLocalBasis` is called twice to compute second order weight derivatives. 

I see two solutions 
- in `addLocalNeighbor` of `MlsSphereFitDer` we can add `evalPos` and then call `DistWeightFunc::spaced2w` (which will substract agin the `evalPos` when calling `convertToLocalBasis`)
- the second order derivatives of the weights are done in the local basis and never call `convertToLocalBasis` (contrary to order 0 and 1)

PR #97, opened 2023-06-21
Fix MlsSphereFitDer weight derivatives (issue #95)

ThibaultLejemble:
Here is a quick fix to the issue  #95.
The basis center is added to the local point to get back the point in global coordinates, and then call weight derivatives functions. 

Small disadvantage: the basis center is added here, but then subtracted in the weight function, so it is not the most efficient solution (except if the compiler is good enough to avoid these successive additions and substractions).

### DIFF
--- a/Ponca/src/Fitting/mlsSphereFitDer.hpp
+++ b/Ponca/src/Fitting/mlsSphereFitDer.hpp
@@ -35,14 +35,14 @@ MlsSphereFitDer<DataPoint, _WFunctor, DiffType, T>::addLocalNeighbor(Scalar w,
         Matrix d2w = Matrix::Zero();
 
         if (Base::isScaleDer())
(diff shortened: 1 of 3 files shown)


nmellado:
Actually the bug can be simply fixed by accessing the global coordinates with `attributes.pos()`.
The documentation should be also fixed.

Ponca/src/Fitting/mlsSphereFitDer.hpp
```diff
@@ -34,15 +34,18 @@ MlsSphereFitDer<DataPoint, _WFunctor, DiffType, T>::addLocalNeighbor(Scalar w,
         // compute weight derivatives
         Matrix d2w = Matrix::Zero();
 
+        // weight derivatives must be computed using global coordinates
+        const VectorType globalQ = Base::m_w.basisCenter() + localQ;
+
```
nmellado:
```suggestion
```

PR #99, opened 2023-07-22
Improved doc in changeBasis() and small helper

julesvidal:
Dear Nicolas,
This PR integrates improvements in the doc of the `changeBasis()` function in `algebraicSphere.h`.
It adds in the function documentation the maths behind the formulas.
Please feel free to ask for modifications in the notations or other.

Additionnally I added a small getter to retrieve the basisCenter of a sphere.
Have a nice summer,
Best,
Jules

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -10,6 +10,7 @@ improved doc.
 - API
     - [spatialPartitioning] Add kd-tree traits type (#80)
     - [fitting] Add Primitive::getNumNeighbors (#86)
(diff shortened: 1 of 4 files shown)


nmellado:
Hello,
Thanks for the proposal.
I made two changes:
 - replace access to basis center by access to the weight function. At some point we will add non-centered basis capabilities, so it is not a good idea to call this method here. Also, I moved the method to `Primitive` instead of `AlgebraicSphere`.
 - Updated the CHANGELOG to keep track of your changes.

Issue #100, opened 2023-07-26
Stable fit with 2 neighbors


ThibaultLejemble:
Hello, 
I found a situation where the fit result is set to `STABLE`, but the number of neighbors is equal to `2`.
It seems to be in contradiction with the documentation that says: `The fitting is stable an ready to use (and having more than 6 neighbours)` 
https://github.com/poncateam/ponca/blob/e24146e1c599debd7ec9a2a6911c0bfe1a7e94ba/Ponca/src/Fitting/enums.h#L16

Is the documentation correct? 
If it is correct I'll check what makes the status as stable.

Here is the example I am using: 

```c++
#include <Ponca/Fitting>
#include <iostream>
#include "Eigen/Eigen"

using namespace Ponca;

class Point
{
public:
  enum {Dim = 3};
  using Scalar = float;
  using VectorType = Eigen::Matrix<Scalar, Dim, 1>;
  using MatrixType = Eigen::Matrix<Scalar, Dim, Dim>;

  inline Point(
    const VectorType& pos,
    const VectorType& normal)
    : m_pos(pos), m_normal(normal) 
    {}

  inline const VectorType& pos()    const {return m_pos;}
  inline       VectorType& pos()          {return m_pos;}
  inline const VectorType& normal() const {return m_normal;}
  inline       VectorType& normal()       {return m_normal;}

protected:
  VectorType m_pos, m_normal;
};

using Scalar = Point::Scalar;
using Vector = Point::VectorType;

using WeightFunc = DistWeightFunc<
  Point,
  SmoothWeightKernel<Scalar>>;

using Fit0 = Ponca::Basket<
  Point, 
  WeightFunc, 
  Ponca::OrientedSphereFit,
  Ponca::GLSParam>;

using FitSphere = Ponca::BasketDiff<
  Fit0, 
  Ponca::FitScaleSpaceDer,
  Ponca::OrientedSphereDer,
  Ponca::MlsSphereFitDer,
  Ponca::CurvatureEstimatorBase,
  Ponca::NormalDerivativesCurvatureEstimator,
  Ponca::GLSDer>;

std::string to_string(FIT_RESULT res)
{
  switch (res)
  {
  case STABLE: return "STABLE";
  case UNSTABLE: return "UNSTABLE";
  case UNDEFINED: return "UNDEFINED";
  case NEED_OTHER_PASS: return "NEED_OTHER_PASS";
[...]

PR #108, opened 2023-08-04
Fix current status (issue #100)

ThibaultLejemble:
This PR mainly fixes issue #100
It also improves
- the doc on `FIT_RESULT`
- the primitive base finalize
- the check on conflict

I have just a few questions or remarks
- the number of neighbors used to determine if a fit is stable or not should depends on the dimension right? For example, a PCA in 2D is stable with 2 neighbors, in 3D with 3 neighbors, etc... I wanted to be sure if it is correct for every fitting procedure and if it is pertinent to add this
- note that fitting an oriented sphere to 1 point and 1 normal vector in any dimension is always stable as it simply returns the plane that passes by this point with a normal equal to the normal of the input point. For 2 points in 2D I believe it is also quite stable in general. That being said, I am not sure for 2 points in 3D, the sphere fit may be good but the derivatives, the GLS, etc, may not be as good, and the stability may depends on the weighting function that is used, so it is safe to keep these thresholds in the end
- the file `linePrimitive.h` has the `Primive` kew-work in its title, whereas other primitives do not have it. For consistency, should we add it to the plane and algebraic sphere file name, or should we remove it from the line file name?

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -10,6 +10,10 @@ Current head (v.1.2 RC)
 
 - Bug-fixes and code improvements
     - [fitting] Use variadic template for basket extensions (#85)
+    - [fitting] Fix current status issue (#108)
+
+-Docs
+    - [fitting] Clarify documentation on FIT_RESULT (#108)
 
 --------------------------------------------------------------------------------
 v.1.1 RC
--- a/Ponca/src/Fitting/covarianceFit.hpp
+++ b/Ponca/src/Fitting/covarianceFit.hpp
@@ -35,11 +35,11 @@ FIT_RESULT
 CovarianceFitBase<DataPoint, _WFunctor, T>::finalize ()
 {
     // handle specific configurations
+    if(Base::finalize() != STABLE) 
+        return Base::m_eCurrentState;
     // With less than 3 neighbors the fitting is undefined
-    if(Base::finalize() != STABLE || Base::getNumNeighbors() < 3)
-    {
+    if(Base::getNumNeighbors() < 3)
         return Base::m_eCurrentState = UNDEFINED;
-    }
 
     // Center the covariance on the centroid
     auto centroid = Base::barycenter();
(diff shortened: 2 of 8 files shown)


nmellado (approved):

nmellado:
> the number of neighbors used to determine if a fit is stable or not should depends on the dimension right? For example, a PCA in 2D is stable with 2 neighbors, in 3D with 3 neighbors, etc... I wanted to be sure if it is correct for every fitting procedure and if it is pertinent to add this

Right, good point. `Dim` is a good threshold.

> note that fitting an oriented sphere to 1 point and 1 normal vector in any dimension is always stable as it simply returns the plane that passes by this point with a normal equal to the normal of the input point. For 2 points in 2D I believe it is also quite stable in general. That being said, I am not sure for 2 points in 3D, the sphere fit may be good but the derivatives, the GLS, etc, may not be as good, and the stability may depends on the weighting function that is used, so it is safe to keep these thresholds in the end

The derivative class (or any other extension) can have a different instability threshold than the main class.

> the file linePrimitive.h has the Primive kew-work in its title, whereas other primitives do not have it. For consistency, should we add it to the plane and algebraic sphere file name, or should we remove it from the line file name?

Yes, right. Another option would be to have Primitive directory, and remove "primitive" in the file name. We could also add a Fit directory, and other files at the root. It would make the code structure more obvious, what do you think ?

ThibaultLejemble:
I will use `Dim` for the neighborhood thresholds that are currently used.
Good idea for the folders `Fit` and `Primitive`, should I add this in this PR, or in another one?

ThibaultLejemble:
For all the sphere fit, the status is `UNDEFINED` with less than `3` neighbors, and `UNSTABLE` with less than `6`.
Do you know why the `6` threshold? 
For now I replaced the `3` by `Dim`, and the `6` by `2*Dim`

nmellado:
@ThibaultLejemble could you please fix the naming issue that makes the tests failing (with `Dim`) ?

> Good idea for the folders Fit and Primitive, should I add this in this PR, or in another one ?

Can be in another one.

nmellado (changes requested):
Seems to be ready for merging (after `Dim` fix).
Please rebase on top of `master`

nmellado (approved):

Issue #107, opened 2023-08-03
Extend requirement system to optional properties


nmellado:
Right now the provide/requirement system is based on enum check, which fail if `Base::PROVIDE_SOMETHING` is not defined.
It does not allow to check the existence of a given properties, and compile different pieces of code depending on the existence (or not) of the property:

```cpp
if (EXISTS(Base::PROVIDE_SOMETHING)) 
   /// do something
else
   /// do something else
```

Any idea on how to do this ?

ThibaultLejemble:
There is one solution to check at compile-time if an enum is defined or not using SFINAE.
Here is one example where a macro `HAS(FEATURE)` is used in `static_assert()`, but also in `if constexpr()` : https://godbolt.org/z/TdToeqPov. It avoids the need to keep up to date a list of all capabilities in `PrimitiveBase` which remains as modular as the current implementation with enums. 

By going further, we can define another macro `PROVIDES(FEATURE)` that declares a new capability `FEATURE` (and assert that it has not already been declared by another base class) : https://godbolt.org/z/d466e5Mnh. The macro `REQUIRES(FEATURE)` is just a `static_assert` that uses `HAS(FEATURE)` and that can replace code like `enum{Check = ...}`.

We could also rely on this mechanism to ensure at compile time that only one class fits a primitive, in replacement to the current dynamic version that use `isValid()` and `CONFLICT_ERROR_FOUND`. Here is one example that defines new macro `FITS(FEATURE)` and `DOES_FIT(FEATURE)` to declare and check that only one class fits a given primitive : https://godbolt.org/z/fxbnrsn9a.

On the other hand, the less flexible solution with the list of all capabilities has its advantages : less unsafe macros, and a global view of what capabilities exist.

nmellado:
@ThibaultLejemble I like the idea to use SFINAE.
Do you think that you could write a first draft ?

I think we could have a macro to declare and collect the capabilities, that could help for the documentation generation for instance.

BDoignies:
@nmellado Done with concepts. Within code:

```
if constexpr (ProvidesAlgebraicSphere<T>)
{
    // Do something
}
```

Concepts are boolean :)

PR #117, opened 2023-11-21
Extend requirement system to optional properties

ThibaultLejemble:
This is a draft for #107. 
The requires/provides system is updated to handle optional capabilities that can be checked at compile-time.
Enums are replaced by `constexpr bool` data members and the verification is done using SFINAE.  

Things that change
- a capability is declared using `PROVIDES(CAPABILITY)`  
- a capability is required by `REQUIRES(CAPABILITY)`
- a macro `IS_PROVIDED(CAPABILITY))` checks at compile-tile if a capability is provided or not by a base class
- enums in the form `PROVIDES_...` are removed (they are now `constexpr bool`, and someone that uses `myfit.PROVIDES_PLANE` might still do the same since `PROVIDES(PLANE)` actually declares a data member named `PROVIDES_PLANE`) 

Note that a class can still override a capability and no error or warning is printed. 
I was thinking about printing at least a warning, but there are some capability such as `NORMAL_DERIVATIVE` that is provided by both `MlsSphereFitDer` and `OrientedSphereDerImpl`, but it is not really a programming mistake, one class overrides the capability and by design inherits the other class. 

Potential next steps
- document all the capabilities and generates the list of classes that provides the capabilities automatically (I am not 100% sure it is possible with doxygen) 
- extend the system with a macro like `FIT(PLANE)` that check if there is no other class that also fits the plane (in replacement to the run-time `isValid()` function) 

Since `enum {Check = PROVIDES_PLANE}` is replaced by `PROVIDES(PLANE)`, I changed all references of `PROVIDES_PLANE` to `PLANE` in the documentation.


nmellado:
Thanks for this very nice proposal.
I suggest to add the detection of duplicated classes in the hierarchy. My suggestion is to add the check in the declaration macro: the base class should not already have the same capability. As this feature is not always wanted (we only want it for container classes), maybe add `PROVIDES_UNIQUE` in addition to `PROVIDES`. 

Does that make sense ?

nmellado:
Another question: could you demonstrate the use of the conditional capabilities usage ? (maybe a test of the macro `IS_PROVIDED`)

ThibaultLejemble:
> Thanks for this very nice proposal. I suggest to add the detection of duplicated classes in the hierarchy. My suggestion is to add the check in the declaration macro: the base class should not already have the same capability. As this feature is not always wanted (we only want it for container classes), maybe add `PROVIDES_UNIQUE` in addition to `PROVIDES`.
> 
> Does that make sense ?

The problem is that it is difficult to know in advance if a capability is unique or not. 
When declaring a new primitive, or a new fitting operation, how to know a priori if some feature won't be overridden by another class, by another user? 
For the current tests to pass, every capabilities can be provided by `PROVIDES_UNIQUE` except `NORMAL_DERIVATIVE` in `OrientedSphereDerImpl` that is overridden by `MlsSphereFitDer`. 
Should I just remove the `PROVIDES(NORMAL_DERIVATIVE)` of `MlsSphereFitDer` (but keeping the method `dNormal`)?

ThibaultLejemble:
> Another question: could you demonstrate the use of the conditional capabilities usage ? (maybe a test of the macro `IS_PROVIDED`)

I added a small test for `IS_PROVIDED` : https://github.com/poncateam/ponca/pull/117/commits/f8900b4980893361214716d8c45aef1dc7d3de15

tests/src/capability.cpp
```diff
@@ -0,0 +1,117 @@
...
+            Base::setPlane(_basisCenter, VectorType::UnitZ());
+        } else {
+            // must not compile if PLANE is provided
+            CHECK<false>();
+        }
+
+        if constexpr( IS_PROVIDED(DOES_NOT_EXIST) ) {
+            // must not compile because DOES_NOT_EXIST does not exist
+            CHECK<false>();
+        }
```
nmellado, line 60: Can't you do something like this ?
```suggestion
        if constexpr( ) {
            // must not compile because DOES_NOT_EXIST does not exist
            CHECK<! IS_PROVIDED(DOES_NOT_EXIST)>();
        }
```

nmellado:
> > Thanks for this very nice proposal. I suggest to add the detection of duplicated classes in the hierarchy. My suggestion is to add the check in the declaration macro: the base class should not already have the same capability. As this feature is not always wanted (we only want it for container classes), maybe add `PROVIDES_UNIQUE` in addition to `PROVIDES`.
> > Does that make sense ?
> 
> The problem is that it is difficult to know in advance if a capability is unique or not. When declaring a new primitive, or a new fitting operation, how to know a priori if some feature won't be overridden by another class, by another user? For the current tests to pass, every capabilities can be provided by `PROVIDES_UNIQUE` except `NORMAL_DERIVATIVE` in `OrientedSphereDerImpl` that is overridden by `MlsSphereFitDer`. Should I just remove the `PROVIDES(NORMAL_DERIVATIVE)` of `MlsSphereFitDer` (but keeping the method `dNormal`)?

We need to take some time to discuss this AFK: some fitting extensions are `Storage` types, while others are `Computation`: some `Computation` object store their results in `Storage` types : this is a side effect, it must be unique in the `Basket`.

tests/src/capability.cpp
ThibaultLejemble: Unfortunately no, because there is a lambda expression in `IS_PROVIDED` and using a lambda-expression in a template-argument is only available with c++20. 
Instead, I can do `static_assert(not IS_PROVIDED(DOES_NOT_EXIST));`, but this piece of code was more to show that the scope after the `if constexpr` is not compiled.

nmellado (commented):
What do you think about highlighting storage-only capabilities, as suggested for curvature ?

nmellado:
Replaced by #293, closing

PR #120, opened 2023-11-29
[SpatialPartitioning] Change the interface of kd-tree nodes

azaleostu:
This PR changes the interface of the kd-tree node to make it more flexible.

Previously, the kd-tree implementation would access node fields directly, which does not allow nodes to customize what data they store and how that data is stored. Now everything is done through methods: the kd-tree calls methods to configure the nodes and the nodes provide accessors to the data the kd-tree and its queries need.

This PR also fixes some small bugs that were introduced with the kd-tree traits.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -8,6 +8,7 @@ Current head (v.1.2 RC)
 - API
     - [spatialPartitioning] Optimize memory use in knngraph queries (#104)
     - [fitting] Mark `Base` type as protected instead of private in CRTP classes (#119)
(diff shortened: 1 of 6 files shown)


nmellado (commented):
Thanks, looks good !
Feel free to mark the PR as ready to review, or update the PR description to add roadmap.

mazestic:
I still need to add some docs and add a proper PR description, I'll mark it ready for review after that

nmellado:
I think this will be the last PR accepted before releasing the version 1.2 (the next PRs will add API changes rooting v2.0)

nmellado (dismissed):

PR #116, opened 2023-10-31
[SpatialPartitioning] Add kd-tree sample index iterator and query

azaleostu:
This PR adds a query that allows iterating over all point samples in a kd-tree.

PR #122, opened 2023-12-05
[SpatialPartitioning] Clean up kdtree API

azaleostu:
This PR removes some unused functions from the kd-tree API, as well as functions that should not be exposed (e.g. non const accessors to internal data).

This PR also fixes the `valid` function of the kd-tree that used to crash immediately upon entry in debug.


azaleostu:
Some of the changes of this PR were taken from #116

nmellado (approved):

PR #127, opened 2023-12-14
Fix sphere gradient (#118)

### LINKED ISSUE #118: Algebraic Sphere gradient is normalized
I think this is an error
```c++
PONCA_MULTIARCH inline VectorType primitiveGradient () const { return m_ul.normalized(); }
```
https://github.com/poncateam/ponca/blob/f8e54a568fe54b9ed78eb612b9ec97101c79df80/Ponca/src/Fitting/algebraicSphere.h#L252C8-L252C8

The gradient of a primitive should be different than the normal vector and should not be normalized?

But without the normalization, the test `gls_paraboloid_der` is broken.
If the gradient should not be normalized, then I'll check what is the issue with this test.

ThibaultLejemble:
Fix #118 
- remove normalization of the gradient of the algebraic sphere in `primitveGradient()`
- the test `gls_paraboloid_der` estimates the derivative of the normal using finite difference on the gradient, so a call to `normalized()` on the gradient is added in this test

### DIFF
--- a/Ponca/src/Fitting/algebraicSphere.h
+++ b/Ponca/src/Fitting/algebraicSphere.h
@@ -249,7 +249,7 @@ public:
     PONCA_MULTIARCH inline VectorType primitiveGradient (const VectorType& _q) const;
 
     /*! \brief Approximation of the scalar field gradient at the evaluation point */
-    PONCA_MULTIARCH inline VectorType primitiveGradient () const { return m_ul.normalized(); }
+    PONCA_MULTIARCH inline VectorType primitiveGradient () const { return m_ul; }
 
     /*!
         \brief Used to know if the fitting result to a plane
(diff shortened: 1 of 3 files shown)


Ponca/src/Fitting/algebraicSphere.h
```diff
@@ -249,7 +249,7 @@ class AlgebraicSphere : public T
     PONCA_MULTIARCH inline VectorType primitiveGradient (const VectorType& _q) const;
 
     /*! \brief Approximation of the scalar field gradient at the evaluation point */
```
nmellado, line 251:
```suggestion
    /*! \brief Approximation of the scalar field gradient at the evaluation point */
        \warning The gradient is not normalized by default
```

But, if we consider the polynomial $u_c + u_l x + u_q x x$, the gradient is supposed to be $u_l + 2u_q x$, right  ?

ThibaultLejemble: Yes, this is what is computed in `primitiveGradient (const VectorType& _q)` with a position as parameter `_q`.

nmellado:
Good for me. Can you please update the changelog ?

nmellado:
Is the `basket` test passing on your computer ?

nmellado (approved):
Test are ok locally, so fine for me to merge.

PR #128, opened 2023-12-14
Simplify KdTree Node type specialization

nmellado:
Add type `KdTreeCustomizableNode` to simplify `KdNode` customization without needed to redefine traits.

Documentation and examples have been added to describe usage.
This PR does not change the client API, so could be included in v.1.2.


azaleostu (approved):
Looks good, it might be getting a bit heavy on templates but I think it's still manageable

PR #135, opened 2024-02-05
Remove exceptions in release mode

nmellado:
This PR remove explicit calls to exceptions in release mode.
It also remove explicit throws that can be easily handled with fallbacks.


azaleostu (approved):
Looks good, thanks!

ThibaultLejemble (approved):

PR #141, opened 2024-07-10
[Fitting] Expose barycenter and add barycenterLocal

nmellado:
In the current version it is not clear that the barycenter is given in local coordinate system.

### Todo
- [x] changelog
- [x] double-check documentation
- [x] methods visibility (local should be for internal use only)

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -5,6 +5,9 @@ Ponca changelog
 --------------------------------------------------------------------------------
 Current head (v.1.3 RC)
 
+- Breaking Changes
+    - [fitting] Fix bad design where public barycenter() function was expressed in local coordinate (#141)
+
 - API
     - [SpatialPartitioning] Change part of the kdtree API (#123)
     - [spatialPartitioning] Refactor KdTree into KdTreeDense + KdTreeSparse (#129)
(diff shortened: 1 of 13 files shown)


nmellado:
This problem has been identified by @wgoadri

dcoeurjo (approved):
Oh I see. Not sure to see the side effects of this edit but the local/global clarification is valuable per se. 

Would that affect some estimators Leo is testing?

azaleostu (approved):
Seems good to me. I agree that the old `barycenter` was unclear whether it was in local or global space. I also agree that code using the old `barycenter` might break, but it should be fine since the change is documented.

nmellado:
It is true that this is a breaking change that will lead to a major release bump (btw we might want to merge it after #143 then).
I checked all internal use. I don't think this will affect downstream code too much, as this function has been introduced recently and is probably not used that much. We will make that clear in the release notes.

nmellado:
I updated the changelog to highlight the breaking change.

nmellado:
Just pushed again to fix a bug when rebasing: I was using old `barycenter()` instead of new `barycenterLocal()` in the tests.

PR #150, opened 2024-07-23
Enable targets exportation during build

### LINKED ISSUE #149: Enable exportation of Ponca during build when linked using add_subdirectory
When Ponca is linked through `add_subdirectory`, exporting a target linked to Ponca would not work because Ponca is not exported during the build.

For example, the following cmake project does not work:

```
cmake_minimum_required(VERSION 3.26)

project(TestPonca)

add_subdirectory(ponca)
add_library(TestPonca STATIC)
target_sources(TestPonca PUBLIC api.cpp api.hpp)
target_link_libraries(TestPonca PUBLIC Ponca::SpatialPartitioning)

install(TARGETS TestPonca EXPORT TestPoncaTargets)
export(EXPORT TestPoncaTargets FILE ${CMAKE_BINARY_DIR}/TestPoncaTargets.cmake)
```

The error:

```
CMake Error in CMakeLists.txt:
  export called with target "TestPonca" which requires target
  "SpatialPartitioning" that is not in any export set.
```

wgoadri:
close #149

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -5,6 +5,10 @@ Ponca changelog
 --------------------------------------------------------------------------------
 Current head (v.1.4 RC)
 
+- Cmake
+    - [build] Enable exportation of projects linking to Ponca targets (#150)
+    - [install] Change output directory to `lib/cmake/Ponca` (#150)
+
 --------------------------------------------------------------------------------
 v.1.3
 This release introduces several improvements around the KdTre API, as well as bug fixes, new features and doc
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -97,7 +97,7 @@ endmacro()
 ################################################################################
 set(Ponca_EXPORT_TARGET_DIR \${CMAKE_CURRENT_LIST_DIR})
 
-set(config_install_dir "${CMAKE_INSTALL_PREFIX}/lib/cmake/")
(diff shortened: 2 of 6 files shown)


wgoadri:
@nmellado @azaleostu I modified the cmake exportation to make the targets available during the build. Now, projects that are exported can link to Ponca. The problem only appeared with libraries.

wgoadri:
I also did a bit of work on the installation folder to add the Ponca name. If Ponca is installed by another project, this makes the install directory more structured. It should not interfere with the previous installation because of the [cmake find_package procedure](https://cmake.org/cmake/help/latest/command/find_package.html#config-mode-search-procedure). 

Also, I think that the installation paths could be shared by the cmake files, they are currently hard coded.

azaleostu (approved):
Maybe update the changelog as this might break existing installs, otherwise looks good 👍

nmellado:
I need to understand why the tests didn't start.

wgoadri:
I think you need to approve the PR to trigger the action because I am a contributor.
To avoid this, the workflow can use the `pull_request_target` event that allows fork from anyone to trigger the test with a pr from a fork. But its not secure. 

https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#controlling-changes-from-forks-to-workflows-in-public-repositories
https://github.com/actions/add-to-project/issues/163

nmellado (approved):

PR #151, opened 2024-08-29
Fix max node count enum

azaleostu:
On my MSVC install, the max node count enum does not automatically get promoted to `std::size_t`, which causes an error as it causes the `MAX_NODE_COUNT` value to be truncated. This PR makes the enum type explicit to avoid this.


nmellado:
Closed on purpose ?

PR #152, opened 2024-10-23
[SpatialPartitioning] Force max node count type to 'std::size_t'

azaleostu:
On MSVC, the `MAX_NODE_COUNT` enum value does not automatically get promoted to `std::size_t` when it is too big which generates an error ([C4309](https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-2-c4309?view=msvc-170)). This PR simply forces the enum to have `std::size_t` as the underlying type.


nmellado (changes requested):
Thanks.
Please update CHANGELOG so we can merge the PR.

nmellado (approved):

PR #153, opened 2025-04-17
[Fitting] Proposal : add transformAndCompute

nmellado:
Add a mechanism to modify neighbors on the fly without modifying the host container.

TODO
- [ ] Add tests


nmellado:
Note relevant anymore : custom basis can be used to transform the neighbors.

PR #156, opened 2025-06-06
[ci] Install pcl on linux for example testing

Me-k-01:


Me-k-01:
Unabling C support seems to solve this issue on my machine. I'm going to revert the downgrading of CMake in the following commits and try this fix on the CI workflow.

nmellado (approved):

Issue #161, opened 2025-06-12
[ci] Random unit test crash


Me-k-01:
Sometimes the continuous test crashes seemingly at random on the GitHub Action workflow.

Some documented cases : 
- On Ubuntu
24/33 Test # 24: fit_line ...........................Subprocess aborted***Exception:   0.50 sec
https://github.com/poncateam/ponca/actions/runs/15443282741/job/43466117498?pr=96

- On MacOS
27/33 Test # 27: basket_2 ...........................Subprocess aborted***Exception:   0.02 sec
https://github.com/poncateam/ponca/actions/runs/15444566025/job/43470563197?pr=96
14/33 Test # 14: fit_radius_curvature_center_7 ......Subprocess aborted***Exception:   0.15 sec
https://github.com/poncateam/ponca/actions/runs/15555330517/job/43816105143?pr=158

- And on Windows
24/33 Test # 24: fit_line ...........................Exit code 0xc0000409
***Exception:   1.44 sec
The link to the crash was lost after a restart

PR #162, opened 2025-06-12
Investigation of the random crash in the unit test

Me-k-01:
This branch is dedicated to fixing the random crash that can happen on the unit test (issue #161 )

Issues fixed:

- [x] #163 
- [x] #170
- [x] #164  
- [x] #171 
- [x] #169 
- [ ] #165

### DIFF
--- a/Ponca/src/Fitting/algebraicSphere.h
+++ b/Ponca/src/Fitting/algebraicSphere.h
@@ -100,7 +100,7 @@ public:
     }
     /// \brief Approximate operator \warning Assume that other shares the same basis \see changeBasis()
     template<typename Other>
-    PONCA_MULTIARCH inline bool isApprox(const Other& other, const Scalar& epsilon = Scalar::dummy_precision()) const{
+    PONCA_MULTIARCH inline bool isApprox(const Other& other, const Scalar& epsilon = Eigen::NumTraits<Scalar>::dummy_precision()) const{
         PONCA_MULTIARCH_STD_MATH(pow);
         const Scalar squaredEpsilon = epsilon*epsilon;
         return  pow(m_uc - other.m_uc, Scalar(2)) < squaredEpsilon &&
--- a/Ponca/src/Fitting/algebraicSphere.h
+++ b/Ponca/src/Fitting/algebraicSphere.h
@@ -98,8 +98,9 @@ public:
                 pow(m_uq - other.m_uq, Scalar(2)) < squaredEpsilon &&
                  m_ul.isApprox(other.m_ul);
     }
-    /// \brief Comparison operator \warning Assume that other shares the same basis \see changeBasis()
-    PONCA_MULTIARCH inline bool isApprox(const AlgebraicSphere<DataPoint, WFunctor, T>& other, const Scalar& epsilon = Scalar::dummy_precision()) const{
+    /// \brief Approximate operator \warning Assume that other shares the same basis \see changeBasis()
+    template<typename Other>
+    PONCA_MULTIARCH inline bool isApprox(const Other& other, const Scalar& epsilon = Scalar::dummy_precision()) const{
         PONCA_MULTIARCH_STD_MATH(pow);
         const Scalar squaredEpsilon = epsilon*epsilon;
         return  pow(m_uc - other.m_uc, Scalar(2)) < squaredEpsilon &&
--- a/Ponca/src/Fitting/algebraicSphere.h
+++ b/Ponca/src/Fitting/algebraicSphere.h
@@ -98,6 +98,15 @@ public:
                 pow(m_uq - other.m_uq, Scalar(2)) < squaredEpsilon &&
                  m_ul.isApprox(other.m_ul);
     }
+    /// \brief Comparison operator \warning Assume that other shares the same basis \see changeBasis()
+    PONCA_MULTIARCH inline bool isApprox(const AlgebraicSphere<DataPoint, WFunctor, T>& other, const Scalar& epsilon = Scalar::dummy_precision()) const{
+        PONCA_MULTIARCH_STD_MATH(pow);
+        const Scalar squaredEpsilon = epsilon*epsilon;
+        return  pow(m_uc - other.m_uc, Scalar(2)) < squaredEpsilon &&
+                pow(m_uq - other.m_uq, Scalar(2)) < squaredEpsilon &&
+                 m_ul.isApprox(other.m_ul);
+    }
+
 
     /*! \brief Comparison operator, convenience function */
     PONCA_MULTIARCH inline bool operator!=(const AlgebraicSphere<DataPoint, WFunctor, T>& other) const{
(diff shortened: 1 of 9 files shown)


nmellado:
Really nice, love it !

Me-k-01:
This will fix the crash of fit_line in the CI

Ponca/src/Fitting/algebraicSphere.h
Me-k-01, line 1: After adding this function to algebraicSphere, I'm getting an error saying that the call to 'isApprox' is ambiguous, in the basket.cpp test. 
Do you have an idea of what is causing the problem? AlgebraicSphere isn't supposed to contain Eigen functions unless I'm mistaken?

nmellado: Hum right, in `Ponca` you can have a Basket with several primitives in it. In this case, calling `isApprox` is ambiguous. 
You might need to refactor the test to add a functor that will first get the right primitive, then call `isApprox`.

nmellado: You already have this with `isSamePlane` and `isSameSphere` (the later can be refactored to use the sphere `isApprox` function instead.
Let's discuss this AFK.

nmellado: Let's skip the hybrid case

tests/src/basket.cpp
```diff
@@ -108,21 +108,20 @@ void testBasicFunctionalities(const KdTree<typename Fit::DataPoint>& tree, typen
...
         fit3.setWeightFunc(WeightFunc(analysisScale));
         fit3.init(fitInitPos);
         fit3.computeWithIds( tree.range_neighbors(fitInitPos, analysisScale), vectorPoints );
         //! [Fit computeWithIds]
-        VERIFY(fit3 == fit3);
-        VERIFY(fit1 == fit3);
-        VERIFY(! (fit1 != fit3));
-
-
+        VERIFY((fit1.isApprox(fit3, epsilon)));
```
Me-k-01, line 124: This test fails on [float](https://github.com/poncateam/ponca/issues/170), [long double](https://github.com/poncateam/ponca/issues/171) and [double](https://github.com/poncateam/ponca/issues/164#issuecomment-2999901680) too. We need a sorting algorithm.

nmellado:
Could you please update the CHANGELOG ?

Ponca/src/Fitting/algebraicSphere.hpp
```diff
@@ -16,16 +16,17 @@ AlgebraicSphere<DataPoint, _WFunctor, T>::project(const VectorType& _q) const
...
-
     Scalar t;
     if(isPlane())
     {
         t = - potential / (norm*norm);
     }
     else
     {
-        t = - (norm - sqrt(norm*norm - Scalar(4) * m_uq * potential)) / (Scalar(2) * m_uq * norm);
+        Scalar disc = norm*norm - Scalar(4) * m_uq * potential;
```
nmellado, line 26: Why this change ?

Ponca/src/Fitting/algebraicSphere.hpp
```diff
@@ -67,6 +68,7 @@ AlgebraicSphere<DataPoint, _WFunctor, T>::potential( const VectorType &_q ) cons
 }
 
 
+
```
nmellado, line 71:
```suggestion
```

Ponca/src/Fitting/algebraicSphere.h
```diff
@@ -262,10 +272,8 @@ class AlgebraicSphere : public T
     PONCA_MULTIARCH inline bool isPlane() const
     {
         PONCA_MULTIARCH_STD_MATH(abs);
-        Scalar epsilon = Eigen::NumTraits<Scalar>::dummy_precision();
-        bool bPlanar   = Eigen::internal::isMuchSmallerThan(abs(m_uq), Scalar(1.), epsilon);
+        bool bPlanar   = m_uq == 0; // Overall, much better than isMuchSmallerThan(abs(m_uq), Scalar(1.), epsilon);
```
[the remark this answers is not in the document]
Me-k-01: Isn't this still going to use the same epsilon value for float and provoke the same loss of precision ? (Which is what caused the bug in the first place)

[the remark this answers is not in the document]
Me-k-01: My mistake, I didn't see the epsilon = 0. Sure this works too.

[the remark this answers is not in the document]
Me-k-01: I checked and yes, this version works the same as m_uq == 0.

Ponca/src/Fitting/algebraicSphere.h
```diff
@@ -262,10 +272,8 @@ class AlgebraicSphere : public T
     PONCA_MULTIARCH inline bool isPlane() const
     {
         PONCA_MULTIARCH_STD_MATH(abs);
-        Scalar epsilon = Eigen::NumTraits<Scalar>::dummy_precision();
-        bool bPlanar   = Eigen::internal::isMuchSmallerThan(abs(m_uq), Scalar(1.), epsilon);
+        bool bPlanar   = Eigen::internal::isApprox(m_uq,Scalar(0));
```
Me-k-01, line 275: The isApprox(m_uq, 0) fixes most of the cases where the projection test fails, but not all of them : There is still one known example where the projection test fails (due to a lack of precision) : When coeff is set to {-2.7, 4.6, -4}, and the seed is 175032073

## Steps to reproduce
- **1 :** Set the seed to 175032073 for the projection test.
- **2 :** In test/src/projection.cpp at line If coeff is set to : 
```cpp
    VectorType coeff = 5 * VectorType::Random();
    coeff[0] = -2.7f;
    coeff[1] = 4.6f;
    coeff[2] = -4.f;
```

We get the following results when launching the projection test on Ubuntu : 
```console
/home/fauberva/CLionProjects/poncascope/external/ponca/cmake-build-release/tests/src/projection s175032073
Initializing random number generator with seed 175032073
Repeating each test 10 times
Test projection for different baskets...
Testing with parabola...
Test ( testFunction<Point, WeightSmoothFunc>() ) failed in /home/fauberva/CLionProjects/poncascope/external/ponca/tests/src/projection.cpp (85)
    std::abs(fit.potential(proj)) < epsilon

Process finished with exit code 134 (interrupted by signal 6:SIGABRT)
```

Me-k-01: In this specific case, the result of std::abs(fit.potential(proj)) can go as high as 0.00309343

Ponca/src/Fitting/algebraicSphere.h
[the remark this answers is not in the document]
nmellado: ok great.

Ponca/src/Fitting/algebraicSphere.h
nmellado: Can you remind me the name of the test ?

nmellado: I pushed a fix that avoid saddle configuration, that are very likely to cause fitting and projection errors.

PR #173, opened 2025-07-03
[Fitting] Proposal : Adding a user-friendly basket fit wrapper

Me-k-01:
This branch goal is to design a wrapper for the Fit computations, that will serve as a user friendly interface for differential quantities computation. It is based on a private fork of Ponca made by @jmaleo and is being readapted to work with the use of macros for simpler definitions.

The user will be able to compute the differential quantities without having to keep track of the Fit, simply by calling the method estimateDifferentialQuantities and providing the name of the fit type, the weight kernel, the radius and the neighborhood of the point.

The fit types that are compiled can be activated or deactivated by changing the ENUM_FITS variable at the cmake building step. Currently, the available values are : "PCA PSS APSS ASO Sphere". By default, the Estimator Factory will compile all the fitting types, but a restricted set of fitting types can be provided by the user in order to reduce the compile time. 

Here is an example on how to change the ENUM_FITS value : 

### In command line
```sh
cmake -S . -B build -DENUM_FITS="PCA APSS ASO Sphere"
```
### In CLion GUI
By going to Settings > CMake > CMake options, and adding the following option : 
`-DENUM_FITS="PCA APSS ASO Sphere"`
Then reload the CMake Project to generate the header containing the new available fitting types.

### DIFF
--- a/Ponca/src/Fitting/Estimators/estimator.h
+++ b/Ponca/src/Fitting/Estimators/estimator.h
@@ -89,12 +89,13 @@ namespace Ponca {
         outputFunc(indexQuery, fit, positionQuery);
         return res;
     }
+
     template <typename DataPoint, typename WeightFunc>
     BasketBase< DataPoint, WeightFunc > getFit(const Estimators::FitType name) {
         switch (name) {
-#define ENUM_FIT(name) \
-            case Estimators::FitType::name :   \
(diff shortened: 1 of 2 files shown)


Me-k-01:
As stated in the description of the last commit, we still need to wait for the PR #175 , so that we can use the BasketBase class to group all the fit type in the factory (in the fit type array). Another possibility would be to simply use a switch, to avoid storing the different fit types into a single array.

Issue #181, opened 2025-07-17
TODO : Review the comments and fix some typos


Me-k-01:
For instance, in the [PR 177](https://github.com/poncateam/ponca/pull/177#discussion_r2209536533), we noted that in the file [Ponca/src/Fitting/weightFunc.h](https://github.com/poncateam/ponca/pull/177/files#diff-ac73d0bac3c32d3598de168f20c4e3037293cda7d93577614039a04852f3563b) at the line  47, the following comment has a typo : "assrt" instead of "assert"
```diff
-         //\todo manage that assrt on __host__ and __device__
+        //\todo manage that assert on __host__ and __device__
```

PR #182, opened 2025-07-17
Fix typo in comment in weightFunc.h  which is line 47  assrt->assert

morty649:


nmellado (dismissed):
Thanks for the contribution.

Fixes #181

nmellado (approved):

PR #187, opened 2025-09-03
[fitting] Adding the MLS computation

Me-k-01:
Adds the MLS process as a module that can be used on the Basket template pattern. A new compute method allows the user to do the MLS iteration on the Fit.

### DIFF
--- a/Ponca/src/Fitting/basket.h
+++ b/Ponca/src/Fitting/basket.h
@@ -196,7 +196,14 @@ namespace internal
             return res;
         }
 
-    private:
+    protected:
+        /*!
+         * \brief Computes the fit using the MLS iteration process.
+         * The position of the projected point is outputted through the lastPosition argument.
+         * \param points An STL-like container of points
+         * \param mlsIter The amount of MLS iteration that is being done for this fit
+         * \return The result of the fit
+         */
         template<typename Func>
         FIT_RESULT computeMLSImpl(Func&& computeFunc, int mlsIter, Scalar epsilon) {
             FIT_RESULT res = UNDEFINED;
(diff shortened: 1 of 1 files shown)


Ponca/src/Fitting/mls.hpp
```diff
@@ -0,0 +1,20 @@
+template<class DataPoint, class _WFunctor, typename T>
+template <typename IndexRange, typename PointContainer>
+PONCA_MULTIARCH inline FIT_RESULT MLS<DataPoint, _WFunctor, T>::computeWithIdsMLS(IndexRange ids, const PointContainer& points){
+    lastPosMLS = Base::getWeightFunc().basisCenter(); // Initial position before projection
+    FIT_RESULT res = FIT_RESULT::UNDEFINED;
+
+    for( int mm = 0; mm < mlsIter; ++mm) {
+        // Starts a new pass and initialise the fit
+        Base::computeWithIds(ids, points); // TODO : Fix this unresolved call
```
Me-k-01, line 9: For some reason, the compiler can't resolve the `Basket::computeWithIds` or `Basket::compute` functions. But it can find the other class methods like `getWeightFunc` or `finalize`. 

I suspect that the `Base` type isn't linked to the `Basket` parent class. I'm not sure how to link it properly since the other examples of **function overwrites** in Ponca do not redefine the compute method, like we want to. 

For instance, in the [Ponca/src/Fitting/gls.h](https://github.com/poncateam/ponca/blob/f7f7d7cc113b4e53be6bee25005fbdbe7e016293/Ponca/src/Fitting/gls.hpp#L11) file, `GLS` redefines the `Basket` methods like `finalize` in the same way and can call the `finalize` of the parent. So I don't know what is different for the `compute` method.

nmellado:
Can be updated according to new master, and merged in 1.4

nmellado (changes requested):

PR #188, opened 2025-09-04
Making Basket inherit ComputeObject

Me-k-01:
To implement MLS (#187)  and CNC (#175), a new compute structure needs to be added. 
This structure will allow us to group all the possible fitting as a computational objects.

### DIFF
--- a/Ponca/src/Fitting/basket.h
+++ b/Ponca/src/Fitting/basket.h
@@ -72,20 +72,49 @@ namespace internal
 }
 #endif
 
+    /*!
+         \brief ComputeObject is a virtual object that represents an algorithm which can be used with the compute functions.
+         The compute(begin, end) and computeWithIds(ids, points) methods must be implemented by the inheriting class.
+         \note The compute(container) that is defined in this structure can be reused in the inheriting class by adding
+         "using ComputeObject<Self>::compute;" to make it accessible
+    */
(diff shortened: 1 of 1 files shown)


nmellado (changes requested):
At some point we will have to update the main documentation to introduce the concept of `ComputeObject`, separated from the fitting Basket type.

Btw I wonder if a better name for `Basket` could not be `FittingBasket`. It is a bit longer, but it conveys the fact that the object is fitted to the data. I think it is the common property of all the objects aggregated in the CRTP. In the same line, the type `Primitive` could be renamed `FitBase` or `FittedObject` or similar (if you have better ideas, feel free to suggest). I guess this PR is the right place to do it, what do you think ?

Me-k-01:
Renaming `Basket` will affect a lot of code and documentation. So I would prefer to keep this PR simple so that we can add it easily to the other previously mentioned PRs that need it. 

I will open another PR instead for this renaming once the discussion #189 is solved.

Me-k-01:
@nmellado Renaming `Basket` will affect a lot of code and documentation. So I would prefer to keep this PR simple so that we can add it easily to the other previously mentioned PRs that need it. 

I will open another PR instead for this renaming once the discussion #189 is solved.

nmellado (approved):

PR #191, opened 2025-09-15
[Fitting] Add features in CovarianceFitBase

azaleostu:
Adds the following features to `CovarianceFitBase`:

- Linearity
- Sphericity
- Anisotropy
- Eigenentropy
- Lambda 0
- Lambda 1
- Lambda 2
- Barycenter distance

### DIFF
--- a/Ponca/src/Fitting/covarianceFit.h
+++ b/Ponca/src/Fitting/covarianceFit.h
@@ -81,31 +81,31 @@ namespace Ponca
         */
         PONCA_MULTIARCH inline Scalar surfaceVariation() const;
 
-        /*! \brief Implements the planarity.
+        /*! \brief Implements the planarity \cite Guinard:2017 .
             Planarity is defined as:
             \f[ \frac{\lambda_1 - \lambda_0}{\lambda_2} \f]
         */
         PONCA_MULTIARCH inline Scalar planarity() const;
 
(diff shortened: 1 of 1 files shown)


Ponca/src/Fitting/covarianceFit.h
```diff
@@ -78,6 +81,53 @@ namespace Ponca
...
+        PONCA_MULTIARCH inline Scalar lambda_1() const;
+
+        /*! \brief The maximun eigenvalue \f$ \lambda_2 \f$.
+        */
+        PONCA_MULTIARCH inline Scalar lambda_2() const;
+
+        /*! \brief The distance between the barycenter and the origin.
+            \return 0 for invalid fits
+        */
+        PONCA_MULTIARCH inline Scalar barycenterDistance() const;
```
nmellado, line 129: Should be defined in the class `MeanPosition`.

nmellado (changes requested):
Looks good, thanks.
Just need to move one method at the proper location + update the CHANGELOG (I can do it if you are not sure how to do it)

azaleostu:
> A general comment regarding the documentation : do you have any reference to scientific papers we could cite for these features ?

I'm not sure which references we have for these features. They were suggested by Loïc Barthe and Adrien Nautre, they might have more information on this.

nmellado (approved):

PR #199, opened 2025-09-22
Update Eigen to 5.0

### LINKED ISSUE #197: Update Eigen to 5.0 RC

nmellado:
Fixes #197 

TODO:

- [x] Fix compilation warnings
- [ ] Fix crash with gcc: with Eigen 5.0.0 we had to disable vectorization to avoid a runtime error in test `basket_2` (with `Scalar=double`). Currently testing with Eigen 5.0.1.
    - [ ] Try to make a minimal example to submit a bug request to Eigen.
- [ ] Be sure that submodule commit id is the same than the one in cmake

### DIFF
--- a/.github/workflows/testAsSubmodule.yml
+++ b/.github/workflows/testAsSubmodule.yml
@@ -38,7 +38,7 @@ jobs:
           git clone https://gitlab.com/libeigen/eigen.git eigen_src
           mkdir eigen_installed/
           cd eigen_src
-          git checkout 3.4.0
+          git checkout 5.0
         if: ${{ matrix.config.installEigen == 'COMPILE'}}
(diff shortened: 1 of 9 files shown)


nmellado:
This version brings a lot of deprecation warning that need to be fixed.
I think we can keep it to Milestone 2.0 as already planned.

BDoignies:
The bug about vectorization is still present (I tried with multiple gcc version and none of them run basket_2 propertly). At least the PR is rebased and I fixed a few Eigen version mismatch. 

The problem of eigen/gcc mismatch is very common it seems. The fix is often to specify the alignment when combining standard library and eigen, though I do not know if we can/should do it here.

PR #214, opened 2025-10-08
Fixing include directives for std lib

### LINKED ISSUE #212: Missing include directive
Sometimes the include directive is missing for the std library (for example `Ponca/src/Fitting/mongePath.h` uses pow but doesn't include cmath). It works for now only because other included file includes those library. This is prone to cause errors, thus we need to make sure that we are including the relevant libraries everywhere we use them.

We also need to ensure we are using the `PONCA_MULTIARCH_INCLUDE_CU_STD` directive every time we are including a module from the std library, to enable support for the CUDA compilation.

Me-k-01:
Fix #212

### DIFF
--- a/Ponca/src/Fitting/mongePatch.hpp
+++ b/Ponca/src/Fitting/mongePatch.hpp
@@ -1,6 +1,7 @@
 ﻿
 #include <Eigen/SVD>
 #include <Eigen/Geometry>
+#include PONCA_MULTIARCH_INCLUDE_CU_STD(cmath)
 
 template < class DataPoint, class _WFunctor, typename T>
 void


Me-k-01:
@nmellado , do we also have to adapt the other objects and functions that are from the std library, for cuda compability? 

For example, we use `std::vector` in `Ponca/src/SpatialPartitioning/KdTree/KdTreeTraits.h` should I also replace it with `PONCA_MULTIARCH_STD_MATH(vector)` and include it through `#include PONCA_MULTIARCH_INCLUDE_STD(vector)`  ?

Here is a complete list of the members of the std library that are currently used in Ponca without using the `MULTI_ARCH` include directive (most of them are from the `Spatial Partitioning Module`) :
- [ ] std::set
- [ ] std::stack
- [ ] std::array
- [ ] std::move
- [ ] std::copy_backward
- [ ] std::is_same
- [ ] std::forward
- [ ] std::ostream
- [ ] std::transform
- [ ] std::back_inserter
- [ ] std::iota
- [ ] std::partition
- [ ] std::make_unsigned
- [ ] std::is_base_of
- [ ] std::distance
- [ ] std::less
- [ ] std::min
- [ ] std::begin
- [ ] std::end
- [ ] std::size_t

PR #220, opened 2025-10-27
[examples] Add a cuda example for 3D point clouds

Me-k-01:
Address #252 by adding an example that does the fitting process of a 3D point cloud, using a CUDA kernel. 

## Changes made by this PR
- [x] Add an example of the fitting process on the GPU for 3D point clouds.
- [x] Change the layout of the CUDA example folder.
- [x] Test the fitting results (output the potential and primitive gradient results of the fit on a plane).
- [x] Add flags for Eigen to work on GPU


Me-k-01:
Splitted the PR in two (#242 will be for the kdtree refactor). 

A clean up of the commits is needed.

Me-k-01:
Ready for review @nmellado

Issue #249, opened 2026-01-23
[examples] `ponca_binding` fails due to buffer overflow error


Me-k-01:
On current master branch, the `examples/cpp/ponca_binding.cpp` example fails.

- With **Linux Ubuntu 24.04** using docker `nvidia/cuda:12.6.0-devel-ubuntu24.04` in debug and release, received output error :
```sh
*** buffer overflow detected ***: terminated
Aborted (core dumped)
```

- With **Linux 6.17.13-1-MANJARO (64-bit)** in debug and release, received output error : 
```sh
malloc(): corrupted top size

Process finished with exit code 134 (interrupted by signal 6:SIGABRT)
```

PR #250, opened 2026-01-26
[examples] Fix `ponca_binding`

### LINKED ISSUE #254: [spatialPartitioning] KdTree's `DefaultConverter` doesn't work with `PointPositionNormalLateBinding`
The default point converter function doesn't work with the `PointPositionNormalLateBinding` Datatype (from PR #220). 

The compiler is trying to convert the PointData type using the `DefaultConverter` but fails. The linter of CLion states the following : 

```sh
std::is_copy_assignable<Ponca::PointPositionNormalLateBinding<float, 3>>::value': type is not assignable
```

And the compiler states the following : 
```sh
/usr/include/c++/10/bits/stl_algobase.h:422:39: error: static assertion failed: type is not assignable
  422 |    static_assert( __assignable::type::value, "type is not assignable" );
```

This is probably caused by the transform code that gets executed at line 160 of the kdTree.h file, because when trying to use a custom converter, the program compiles and run normally, e.g. :

```cpp
KdTreeDense<DataPointBinding> tree;
tree.build(std::forward<std::vector<DataPointBinding>>(points), [](auto&& inputPointContainer, auto& internalPointContainer) {
        internalPointContainer = std::forward<decltype(inputPointContainer)>(inputPointContainer);
});
```

So we can assume that the expression `constexpr (std::is_same<InputContainer, PointContainer>::value)` is false for this DataType, even through it shouldn't be.

Me-k-01:
Fix `ponca_binding` example by solving bad memory management.

# Changes made by this PR

- [x] Resolve issue #249
- [x] Add a binding datatype test to the CI to prevent regression
- [x] Add a `is_copy_assignable_v` check in the KdTree DefaultConverter function to make the KdTree compatible with the alternative Point Binding Data Types introduced by this PR (Fix #254)

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -16,10 +16,12 @@ Current head (v.2.0.alpha1 RC)
 - Tests
     - [all] Speed up tests in coverage mode (#236)
     - [common] Use common point type and point generation methods (#248, #220)
+    - [ci] Add binding datatype test to prevent regression (#250)
 
 - Examples
     - [nanoflann] Fix broken include directives (#236)
     - [common] Use common point type and point generation methods (#248)
+    - [examples] Fix `ponca_binding` buffer overflow error (#250)
 
 - CI/Actions
     - Add coverage report for Pull Requests (#236, #245, #245)


Me-k-01:
@nmellado Ready for review
/!\ Needs to be rebased with master once #220 is merged /!\

Me-k-01:
This PR is now up to date with master, and the suggested changes were made.

It is now ready for review @nmellado.

PR #90, opened 2023-06-13
Add RimlsPlaneFit

AntoineLafon:
This PR add a new fitting method based on the RIMLS.

PR #271, opened 2026-03-10
RIMLS Basket compute function

BDoignies:
Hi, 

This PR proposes adding the RIMLS [[Ozterili 2009]](https://inria.hal.science/inria-00354969/PDF/RIMLS_eg09.pdf) general framework within a Basket::compute function (as discussed with @nmellado, unlike #90).

There are a few decisions I have made that should be discussed:
- Gradient estimation: the method requires a gradient estimate initially (if I have understood correctly). The Basket API does not offer this in the general setting. Should we move the code inside the BasketDiff function if it provides a uniform API? For now, I propose relying on an estimate which is essentially the directional derivative of the SDF in the direction of the projected point. Another option would be to use proper finite differences, but choosing the 'step' could be tricky in the general setting due to stability issues and could involve more computing, though I presume this would be negligible for simple primitives. 
- The termination criterion for re-fitting the SDF was changed to reduce computational cost (is this the right term?). The article suggests using a relative difference in weighting terms per neighbour, which requires storage (that could be a problem for CUDA kernel) and processing time. Instead, I suggest using an absolute difference of the values, which is way simpler to compute.

I also have an important question: where should the result be stored? MLS chose to store the result within the evalPos. Assuming there is no eigen aliasing, and if this is the correct place to do so, then nothing should be changed in the code (but removing TODO). 

I am also unsure about testing within the library. I copied the ones for MLS regression, which seems fine. But perhaps we should expect more from RIMLS?


nmellado (commented):
As discussed AFK, might make sense to move the computeXXX variants (not the compute and computeWithIds) as namespace functions.
Also, regarding the naming conventions, `compute` is associated to the fitting itself, while the computeXXX describe projection operators that use compute internally, so we need to reflet this somehow.
Also, in the current version of the Fitting package, each file stand for a class (except utils files). If we had function namespace, we need to define a convention for the files storing them.

PR #274, opened 2026-03-12
[common] Clang format for Ponca

BDoignies:
Hi, 

Here is my initial guess at creating a standard formatter for  Ponca. I suggest using Clang Format, one of the most customisable formatters available. 

@nmellado and @Me-k-01, please take a look at the current commit and let me know if there are any lines that don't look right to you. 

To test a new configuration for the entire repository: "find Ponca/ | xargs clang-format -i".
Alternatively, you can use the online tool at https://zed0.co.uk/clang-format-configurator/ (upload the .clang-format file) to quickly view the rules. Be mindful of spaces in and out of parentheses, templates, etc. There are plenty of alignment options, but bear in mind that they are enforced by the formatter. For example, if you select the option to align consecutive assignments, they will always be aligned, no matter what. 

Send me back the new formatter and I will rewrite the commits so that this is easier to see global changes.

Also, the old guidelines state that parameters should start with '_', but there is no rule to enforce this within Clang Format. However, we could also add a Clang Tidy check that can do this (and much more).

Once the formatter is set, we can add a GitHub action check to maintain consistency, but this may cause more problems than it solves.

### DIFF
--- a/Ponca/src/Fitting/cnc.h
+++ b/Ponca/src/Fitting/cnc.h
@@ -56,8 +56,8 @@ public:
 
     /*! \brief Get the position of the point at the given index.
      *
-     * @param index Index of one of the three vertices of the triangle (between 0 and 2)
-     * @return The position of vertex
+     * \param index Index of one of the three vertices of the triangle (between 0 and 2)
+     * \return The position of vertex
      */
     PONCA_MULTIARCH [[nodiscard]] VectorType& getPos(const int index) {
         return m_points[index];
--- a/Ponca/src/Fitting/cnc.h
+++ b/Ponca/src/Fitting/cnc.h
@@ -14,249 +14,241 @@ All rights reserved.
 #include "weightFunc.h"
 #include "weightKernel.h"
 
-
 namespace Ponca
 {
 
-namespace internal {
-/*!
- * Stores the three points and normals of the triangles and provides access to Corrected Normal Current formula
- *
- * \tparam DataPoint Type of input points.
- *
- * \see CNCEigen
- */
-template < class DataPoint >
-struct Triangle {
-public:
-    typedef typename DataPoint::Scalar Scalar;
-    typedef typename DataPoint::VectorType VectorType;
-    typedef typename DataPoint::MatrixType MatrixType;
-protected:
-    std::array < VectorType, 3 > m_points;
-    std::array < VectorType, 3 > m_normals;
-public:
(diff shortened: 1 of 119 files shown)


Ponca/src/Fitting/cnc.h
```diff
@@ -9,156 +9,165 @@ All rights reserved.
...
  * - The principal curvatures values and directions
  * - The mean and gaussian curvatures
  *
  * \see PROVIDES_PRINCIPAL_CURVATURES
-*/
-template < class P, TriangleGenerationMethod _method = UniformGeneration>
-class CNC : ComputeObject<CNC<P, _method>> {
-protected:
+ */
+template<class P, TriangleGenerationMethod _method = UniformGeneration> class CNC : ComputeObject<CNC<P, _method>>
```
Me-k-01: I'm not a fan of removing the end of line in between templates and the class / function name typically because of those kinds of examples. Are we sure we want to put everything on the same line? It can look a bit crowded in some places... I find it easier to read when we have a dedicated template line. 

Is there a way to avoid this issue with the formatting tool?

Me-k-01 (changes requested):
The new formatting looks overall good, however I would advice against the removal of the end of line after templates arguments, because it makes the templated code less readable. 

I've pointed out an [example in my review](https://github.com/poncateam/ponca/pull/274#discussion_r2924979986) where the lines are way too long if we put the templated arguments on the same line as the function or class name. 

Is there a way to roll back this change with the new formatting tool ?

BDoignies:
@nmellado @Me-k-01 New version of the formatter is pushed (+ a systematic replace of doxygen @ comments). 

A few questions regarding formatter:
- I enforced indentation with macros as this was quite common in Ponca. However, the formatter allows for the two following styles:
(Current formatting)
```c++
#ifdef WHATEVER
    #define SOMETHING
#endif
``` 
OR
```c++
#ifdef WHATEVER
#    define SOMETHING
#endif
```
They both appear in the library. 

- Whether template has a space after it:
 (Current formatting) 
```c++
template<typename T>
```
OR
```c++
template <typename T>
```

Again both are common in Ponca. 

Also, the maximum line length has been set to 120 for now (I think this is the default for Microsoft style which is the base used). Any other number is fine. Just let me know if any lines have been erroneously broken and if this limit is too low for Ponca.

nmellado (approved):

PR #219, opened 2025-10-20
[build] Update to cpp 20

Me-k-01:
In this PR, the project was updated from cpp 17 to version 20 in order to use concepts

We currently have multiple problems when trying to upgrade the cpp language version on the CUDA examples. A summary of the issues is available in the **troubleshooting notes** section

# Changes made by this PR
- [x] Update CMake to use CMAKE_CXX_STANDARD 20
- [ ] Also update to cpp 20 on the CUDA examples
- [ ] Replace the PROVIDES enum with concepts (to fix the warnings on the PROVIDES enum)
- [ ] Check with CMake that nvcc>=12 before building

# Troubleshooting notes
- **I.** Moving to CPP 20 causes many warnings with the enum PROVIDES system currently in place in the Fitting module (because of an enum to int conversion). To clean up the warnings, we want to replace this with the **concepts** feature of CPP 20.

- **II.** When building the CUDA examples with NVIDIA toolkit 11.8.89 (through docker image `nvidia/cuda:11.8.0-devel-ubuntu22.0`)  we receive the following error:
```sh
CMake Error in examples/cuda/CMakeLists.txt:
  Target "cuda_fit_plane" requires the language dialect "CUDA20" , but CMake                                                                                                                                                   
  does not know the compile flags to use to enable it.                                                                                                                                                                         
                                                                                                                                                                                                                               
                                                                                                                                                                                                                               
CMake Error in examples/cuda/CMakeLists.txt:
  Target "cuda_fit_plane_kdtree" requires the language dialect "CUDA20" , but                                                                                                                                                  
  CMake does not know the compile flags to use to enable it.                                                                                                                                                                   
                                                                                                                                                                                                                               
                                                                                                                                                                                                                               
CMake Error in examples/cuda/ssgls/CMakeLists.txt:
  Target "ponca_ssgls" requires the language dialect "CUDA20" , but CMake                                                                                                                                                      
  does not know the compile flags to use to enable it.                                                                                                                                                                         
```
This is caused by the CMakeLists.txt of the CUDA example : `set(CMAKE_CUDA_STANDARD 20)` seems to trigger this error, but only when we use CPP 20 and not CPP 17 ([see related post](https://stackoverflow.com/questions/75010206/cuda-12-c20-support-with-cmake-not-working)) . Moving to NVCC 12 should solve this issue, but we should add a check for the nvcc version in CMake.

[...]

PR #293, opened 2026-03-24
[common] C++20 Concepts for ponca

### LINKED ISSUE #179: add bases classes for centered/non centered fit

### LINKED ISSUE #218: PROVIDES_PRINCIPAL_CURVATURES is missing from the doc

BDoignies:
Hi, 

This PR serves as a base for introducing concepts in the library. Note that it depends on #219 but some commits will drop once rebase. 

This first commit introduces:
- A format for requires clause
- An example concept for points. Note that due to PointBinding type, we can not check for return type, for write access or for a non const version. 

I suggest to start concept by the prefix "Is" or "Provides" because of current Ponca terminology.


nmellado:
I gave it a go and modified the current proposition. I also modified the curvature estimator, as I believe it is an important feature. 
I did not tried to solve the issues you mentioned in the comments for the missing concepts, will work on it later.
We also need to work on the documentation, I put some comments here and there, but it can definitely be improved.

nmellado:
@BDoignies I finished my first round of changes. Could you please have a look and summarize issues/leftovers ?

BDoignies:
@nmellado 

I added concepts related to filters and frames. In my mind, Frame = local basis and Filter = Frame + distance-based weighting. As far as filters are concerned, I think we can add a few more regarding derivation (probably for BasketUnit and BasketUnitDer, but not for other extensions).

I also split the notion of frame center and evaluation location. To me, these are very different things; at the very least, there should be a semantic distinction (i.e., having both center and evalLocation, even if they are implemented by the same function). This may also allow global frames to take no parameters in their constructor. If we like this change, we should update all code from:

```cpp
Fit f;

for (auto p : points)
{
     f.setNeighborFilter({p.pos(), t});
      // compute, ...
}
```

To something like (exact API to be defined):
```cpp
Fit f;
f.getNeighborFilter().setScale(t); // And other params
// Or we can set only once the filter

for (auto p : points)
{ 
      f.setEvalPoint(p); // Sets internal frame center (as this is a point, it can also automatically set the normal if needed)
      // compute, ...
}
```

This allows for a more uniform API overall. I am especially concerned about GlobalFilter having a reference position and scale in its constructor, as well as NeighborFilterStoreNormal, where the normal is not set. Perhaps we should think a bit more about these concepts.

Ponca/src/Common/pointTypes.h
```diff
@@ -98,6 +98,11 @@ namespace Ponca
         using VectorType = Eigen::Matrix<Scalar, Dim, 1>;
         using MatrixType = Eigen::Matrix<Scalar, Dim, Dim>;
 
+        PONCA_MULTIARCH inline PointPositionNormalBinding()
```
nmellado: The map is stored as const, so it cannot be changed. Why do we want to be able to create ill-formed points ?

Ponca/src/Common/pointTypes.h
```diff
@@ -137,6 +142,8 @@ namespace Ponca
         using VectorType = Eigen::Matrix<Scalar, Dim, 1>;
         using MatrixType = Eigen::Matrix<Scalar, Dim, Dim>;
 
+        PONCA_MULTIARCH inline PointPositionNormalLazyBinding() : m_interlacedArray(nullptr), m_id(0) {}
```
nmellado: Same here, I don't think we want this default constructor.

Ponca/src/Fitting/basketUnit.h
```diff
@@ -46,6 +40,10 @@ namespace Ponca
         Scalar m_sumW{0};
 
     protected:
+        //! \brief Eval location
+        //! \warning This might be different from frame().center() !
+        DataPoint m_evalPoint;
```
nmellado: Let's discuss this AFK,

Ponca/src/Fitting/Basket/fit/mongePatch.h
```diff
@@ -29,33 +30,25 @@ namespace Ponca
...
 
         This primitive provides:
-        \verbatim PROVIDES_MONGE_PATCH, PROVIDES_FIRST_FUNDAMENTAL_FORM_COMPONENTS,
-       PROVIDES_SECOND_FUNDAMENTAL_FORM_COMPONENTS \endverbatim
+        \verbatim PROVIDES_MONGE_PATCH, ProvidesFirstFondamentalFormComponents,
+       ProvidesSecondFondamentalFormComponents\endverbatim
 
         This primitive requires:
-        \verbatim PROVIDES_PLANE, PROVIDES_TANGENT_PLANE_BASIS, PROVIDES_HEIGHTFIELD \endverbatim
+        \verbatim ProvidesPlane, ProvidesTangentPlaneBasis, PROVIDES_HEIGHTFIELD \endverbatim
```
nmellado, line 37: Fix provide list

nmellado (commented):
Minor comments.

Ponca/src/Fitting/Basket/fit/mongePatch.h
BDoignies: Wdym? I think this is best if it is linked to the concept (ie ProvidesPlane) rather than something that only exists within the documentation like "PROVIDES_PLANE"

BDoignies:
@nmellado Does the name suits you "FilterWithAttributes" ? I had other thing in mine but this one was the best IMHO : 

* FilterWithMetadata : this is not really metadata
* FilterWithData: Not really clear if data relates to DataPoint, or somehing else
* (Meta)DataFilter: Looks like we are filtering the (meta)data
* DataStoreFilter: Looks like we are filtering a "DataStore"
* AnnotatedFilter: I really like this one, just don't know if annotated is the right term 
* WrappedFilter: Feels like it does things other than just storing data

nmellado:
Hello @BDoignies ,
FilterWithAttribute is good.

PR #298, opened 2026-03-26
[fitting] Compute functions

BDoignies:
Hi, 

As discussed in #271, this PR moves compute MLS function in a separate file name 'computeScheme.h', because 'compute' is already taken ; that could be change easily. 

A key change is that computeMLS iteration count is no longer defaulted for some overloads. For some reason I do not understand, the compiler calls the wrong overload at this line : [msl:69](https://github.com/poncateam/ponca/blob/1ffd926693113b7db54c52b90e144fa17574978f/tests/src/mls.cpp#L68) (I though SNIFAE would apply but maybe it is not substitution failure afterall)... I do not think this is a bad thing but this is a slight change of API that needs consideration in my opinion.
 
As a side note, previous version had no problem because there was only a single overload possible: computeMLS with a container. I added a new one from two iterator to mimic what compute can be called with.


BDoignies:
Now merged with #271

BDoignies:
@nmellado as discussed AFK, I removed RIMLS from this PR and added projection classes

PR #301, opened 2026-03-31
[spatialPartitioning] Adapt KnnGraph for CUDA

### LINKED ISSUE #263: Extend the KnnGraph API to work on the CUDA Kernel

### LINKED ISSUE #307: [common] Bug in  `limited_priority_queue::push`
# `std::move` is called after the value was already moved

The `std::move` directive is repeated two times on the same value (line 342 and line 347). This could lead to undefined behaviors, as the value may already be in a moved-from state.

https://github.com/poncateam/ponca/blob/73247abfe24d29406a95aee1d4dfa2d34da85d4c/Ponca/src/Common/Containers/limitedPriorityQueue.h#L342-L347

We should wait to know where to insert before moving the value.

# Duplicated code
The `bool limited_priority_queue<T, Cmp>::push` code also has duplicated logic. We should factorize it into one function that accounts for both lvalue and rvalue, using `std::foward` instead of `std::move`

Me-k-01:
Fix #263 by adapting the KnnGraph structure for CUDA

# Changes made by this PR

- [x] Add a `StaticKnnGraph` class and a `KnnGraphPointerTraits` struct for compilation on the GPU
- [x] Add a `Buffers` structure internal to the KnnGraph that can be sent to the GPU
- [x] Add an example of the KnnGraph being transfered inside the CUDA kernel that uses the same logic from the kdtree example
- [x] Update limited_priority_queue to work in CUDA
- [x] Update KdTree k-nearest-neighbors query to work in CUDA by adding a fixed max neighbors length
- [x] Update KnnGraph k-nearest-neighbors query to work in CUDA by adding a fixed max neighbors length
- [x] Replace query containers with custom Stack and HashSet structure
- [x] Added a Bitset variant (faster run time at the cost of more memory usage)  
- [x] Add KnnQuery pointer test in CI for kNearestNeighbors query
- [x] Add KnnQuery pointer test in CI for rangeNeighbors query
- [x] Update KnnGraph range-neighbor query to work in CUDA
- [x] Add unit test for BitSet and HashSet on CI


Ponca/src/SpatialPartitioning/KnnGraph/Query/knnGraphKNearestQuery.h
```diff
@@ -50,19 +50,25 @@ namespace Ponca
...
+        {
+            return std::begin(m_graph->samples()) + QueryType::input() * m_graph->k();
+        }
 
         /// \brief Returns an iterator to the end of the k-nearest neighbors query.
-        inline Iterator end() const { return m_graph->index_data().begin() + (QueryType::input() + 1) * m_graph->k(); }
+        PONCA_MULTIARCH inline Iterator end() const
+        {
+            return std::begin(m_graph->samples()) + (QueryType::input() + 1) * m_graph->k();
+        }
```
Me-k-01: The iterators returned by `KnnGraphKNearestQuery::begin` and `KnnGraphKNearestQuery::end` needs to be readapted for the case where we are using raw data pointers for CUDA. 

We currently use the begin function from `m_graph->samples`. Changing this to `std::begin` won't solve the issue since `std::begin` is not available on the CUDA kernel. 

Other Queries didn't have this issue since they use limited_priority_queue or custom kdtree traversal functions.

Me-k-01: @nmellado In summary we have `m_graph->samples` that can either be of type `std::vector` or `int*` (on the GPU). Do we need to dispatch the the type of the Index container? Or do you see a way to reconcile Pointer arithmetic with STL-like containers?

nmellado: functions taking iterators can take raw pointers, e.g
```cpp
#define SCALAR float
#define DIM 2
#define NBPOINTS 50
#define SCALE float(0.3)
using Point   = Ponca::PointPositionNormal<SCALAR, DIM>;

Point data[NBPOINTS];

// fill data with values
...
// compute fit
const auto q = Vector::Random().eval();

b.setNeighborFilter(NFilter(q, SCALE));
b.compute(data, data+NBPOINTS);
```

so I guess we can find a solution.

Me-k-01:
After further testing, it seems I was wrong about the limited priority queue working on CUDA. It's using `std::vector` which isn't available. Therefore the `kdtree.kNearestNeighbors` query shouldn't work either on CUDA (since it uses the queue structure). 

I'll add a TODO and a warnings on the SpatialPartitioning methods that aren't working on CUDA currently.

Ponca/src/SpatialPartitioning/KnnGraph/Query/knnGraphKNearestQuery.h
Me-k-01: The Knearest Query doesn't work either on the kdtree because it uses an `std::vector` in the queue structure. I'll mark the two `kNearestNeighbors` methods as host only for now.

Me-k-01:
There is still some warnings left to fix in CUDA that are triggered when calling `kdtree.kNearestNeighbors` (about host function being used in host device functions).

Me-k-01:
The warnings are triggered when calling `kdtree.kNearestNeighbors` in a CUDA environment :

```sh 
/CLionProjects/poncascope/external/ponca/Ponca/src/SpatialPartitioning/KdTree/Query/../../query.h(151): warning #20011-D: calling a __host__ function("Ponca::QueryOutputIsKNearest<int, float> ::QueryOutputIsKNearest(int)") from a __host__ __device__ function("Ponca::QueryOutputIsKNearest<int, float> ::QueryOutputIsKNearest [subobject]") is not allowed
```

This is caused by a mismatch of host / device functions : We changed the `QueryOutputIsKNearest` constructor to be host-only (because it's using the `limited_priority_queue` datastructure) while still relying on `QueryOutputBase` as a parent, which then still defines host / device constructor. To solve this warning, we would have to either : 
- Change the way Queries are constructed to allow for host only queries
- Or adapt the `limited_priority_queue` so that it can be compiled on the device. The internal container of the `limited_priority_queue` doesn't seem to be resized anywhere in the code, so we might get away with using a structure that has an internal array with a length fixed at compile time instead, which is the requirement to be able to be instantiated inside a CUDA kernel.

examples/cuda/cuda_utils.cu
```diff
@@ -32,30 +33,32 @@
  * \param deviceBuffers The pointer to the Buffers structure on the device
  * \see freeBuffersOnDevice to free memory on the device with hostBuffersHoldingDevicePointers as an argument
  */
-template <typename Traits, typename KdTreeDenseBuffers, typename StaticKdTreeBuffers>
+template <typename Traits, bool SendNode = true, typename KdTreeDenseBuffers, typename StaticKdTreeBuffers>
```
nmellado: Need to add a documentation for the template parameter `SendNode`.

examples/cuda/ponca_fit_knngraph.cu
```diff
@@ -0,0 +1,193 @@
...
+    //! [Create KdTree on the GPU]
+
+    // Set up the fit
+    Fit fit;
+    VectorType pos = buffers->points[i].pos();
+    fit.setNeighborFilter({ pos, analysisScale });
+
+    //! [Use KdTree on the GPU]
+    // fit.computeWithIds(knngraph.rangeNeighbors(i, analysisScale), knngraph.points());
+    fit.compute(buffers->points, buffers->points + buffers->points_size);
```
nmellado: I think we need to solve this to have the PR well formed. Let's discuss it AFK.

examples/cuda/cuda_utils.cu
Me-k-01: Some documentation about this tparam

Me-k-01:
The KdTree and the KnnGraph now fully works in CUDA. There are still some warnings left that are provoked outside the CUDA kernel, during the building phase of the KnnGraph. I'm still a bit unsure why those warnings aren't happening during the KdTree building process too, since we are doing similar things there, which should also trigger the same warnings. 

But aside from that, the every KnnGraph / KdTree queries works in CUDA now. So I believe we can finally start the review process @nmellado and @BDoignies

examples/cuda/ponca_fit_knngraph.cu
```diff
@@ -0,0 +1,167 @@
...
+    for(unsigned int i = 0; i < nbPoints; ++i) {
+        points[i] = Ponca::getPointOnPlane<DataPoint>(
+            center, direction, width,
+            _bAddPositionNoise, _bAddNormalNoise, _bUnoriented
+        );
+    }
+
+    //! [Build KnnGraph for CPU]
+    Ponca::KdTreeDense<DataPoint> kdtree(points);
+    Ponca::KnnGraph<DataPoint>    knngraph(kdtree, points.size()-1);
```
nmellado: `k` is usually quite small. Why a so large value ?

Ponca/src/Common/Containers/bitset.h
```diff
@@ -0,0 +1,138 @@
...
+    template <int N, typename T>
+    BitSet<N, T>::BitSet() = default;
+
+    template <int N, typename T>
+    PONCA_MULTIARCH void BitSet<N, T>::clear()
+    {
+        for (int i = 0; i < ARRAY_SIZE; i++)
+        {
+            m_data[i] = T(0);
+        }
```
nmellado: Maybe fallback to std::fill when not on cuda ?

Ponca/src/Common/Containers/bitset.h
```diff
@@ -0,0 +1,138 @@
...
+    }
+
```
nmellado: Add comments (especially regarding the returned value)

Ponca/src/Common/Containers/bitset.h
```diff
@@ -0,0 +1,138 @@
...
+                         "Attempted to remove a value that is outside the scope of the BitSet");
+        const int byte          = value / BIT_SIZE;
+        const int bit           = value % BIT_SIZE;
+        const T bitMask         = (T(1) << bit);
+        const bool alreadyEmpty = (m_data[byte] & bitMask) != 0;
+        m_data[byte] &= ~bitMask;
+        return alreadyEmpty;
+    }
+
+    template <int N, typename T>
```
nmellado: Add comments

Ponca/src/Common/Containers/hashset.h
```diff
@@ -0,0 +1,138 @@
...
+            PONCA_ASSERT_MSG(x >= 0 && x < N, "Index is outside the scope of the HashSet");
+            return (x * 2654435761u) % N;
+        }
+    };
+
+    template <int N, typename T>
+    PONCA_MULTIARCH void HashSet<N, T>::clear()
+    {
+        for (int i = 0; i < N; i++)
+            table[i] = EMPTY;
```
nmellado: Fallback to std::fill when not compiled on cuda

Ponca/src/Common/Containers/limitedPriorityQueue.h
```diff
@@ -116,235 +126,190 @@ namespace Ponca
...
+        size_t m_capacity{0}; //!< The capacity of the limited queue
     };
 
     ////////////////////////////////////////////////////////////////////////////////
     ////////////////////////////////////////////////////////////////////////////////
     ////////////////////////////////////////////////////////////////////////////////
     ////////////////////////////////////////////////////////////////////////////////
 
-    // limited_priority_queue ------------------------------------------------------
+    // LimitedPriorityQueue ------------------------------------------------------
```
nmellado, line 160: Makes sense to move to .hpp file maybe

Ponca/src/SpatialPartitioning/KdTree/Query/kdTreeKNearestQueries.h
```diff
@@ -29,8 +29,8 @@ namespace Ponca
         using Scalar         = typename DataPoint::Scalar;
         using VectorType     = typename DataPoint::VectorType;
         using QueryAccelType = KdTreeQuery<Traits>;
-        using Iterator       = IteratorType<typename Traits::IndexType, typename Traits::DataPoint>;
-        using Self           = KdTreeKNearestQueryBase<Traits, IteratorType, QueryType>;
+        using Iterator = IteratorType<typename Traits::IndexType, typename Traits::DataPoint, Traits::MAX_KNN_SIZE>;
+        using Base     = KdTreeKNearestQueryBase<Traits, IteratorType, QueryType>;
```
nmellado: Why this change ?

Ponca/src/SpatialPartitioning/KdTree/kdTree.h
```diff
@@ -277,25 +277,25 @@ namespace Ponca
         /// The returned object can be reset and reused with the () operator
         /// (using the same argument types as parameters).
         ///
-        /// \param point Point from where the query is evaluated
-        /// \param k Number of neighbors returned
-        /// \return The \ref KdTreeKNearestIndexQuery mutable object to iterate over the search results.
-        /// \see KdTreeKNearestQueryBase
+        /// \warning This method can't be called in a CUDA kernel because KdTreeKNearestPointQuery relies on STL-like
+        /// vectors. \param point Point from where the query is evaluated \param k Number of neighbors returned \return
```
nmellado: Still true ?

Ponca/src/SpatialPartitioning/KdTree/kdTreeTraits.h
```diff
@@ -357,7 +358,8 @@ namespace Ponca
             /*!
              * \brief A compile-time constant specifying the maximum depth of the kd-tree.
              */
-            MAX_DEPTH = 32,
+            MAX_DEPTH    = 32,
+            MAX_KNN_SIZE = 10000 //!< The maximum size of a knn query
```
nmellado: That's quite large. Did you check the impact on the performances and memory occupancy ?  This might affect cuda kernels as it will require more memory per thread, and thus increase cache misses. I think that something like 128 is already a good value.

Ponca/src/SpatialPartitioning/KnnGraph/knnGraphTraits.h
```diff
@@ -17,6 +17,10 @@ namespace Ponca
     template <typename _DataPoint>
     struct KnnGraphDefaultTraits
     {
+        enum
+        {
+            MAX_KNN_SIZE = 10000 //!< The maximum size of a knn query
```
nmellado: Same here, seems quite large.

Ponca/src/SpatialPartitioning/KnnGraph/knnGraphTraits.h
```diff
@@ -45,4 +49,43 @@ namespace Ponca
...
+    /*!
+     * \brief Variant to the KnnGraph Traits type that uses pointers as internal storage instead of an STL-like
+     * container.
+     */
+    template <typename _DataPoint>
+    struct KnnGraphPointerTraits
+    {
+        enum
+        {
+            MAX_KNN_SIZE = 10000 //!< The maximum size of a knn query
```
nmellado: Same here, seems quite large

tests/common/kdtree_utils.h
```diff
@@ -328,4 +327,3 @@ std::chrono::milliseconds testQuery(PointContainer& points, MutableQueryFunctor
     }
     return std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::system_clock::now() - time);
 }
-
```
nmellado, line 331: Missing line at the end of file

Ponca/src/Common/Containers/bitset.h
Me-k-01: Everything is already commented at the function declaration (see line 53)

Ponca/src/Common/Containers/bitset.h
Me-k-01: Same here, it's already commented at line 62

Ponca/src/Common/Containers/bitset.h
Me-k-01: We currently aren't using the clear function in PONCA : the Set is just used once to verify if the element was already added or not into the Stack of the `KnnGraphRangeNeighborsQuery`. 

Instead of splitting the code like that (which could introduce bugs in the future as we don't have a way to autocratically test the CUDA version in the CI), we could just simply remove this function, as it was only added in case we ever need to clear a BitSet in the future (for extended compatibilities with set-like interface).

This is also the case of the `BitSet::find`, `BitSet::search` and the `BitSet::flip` method, all of them are untested by the CI and where only added for legacy purposes. 

The only method we currently use is the `BitSet::insert`, which also does a kind of search, as we return false if the element was already inserted

What do you think @nmellado ?

Ponca/src/Common/Containers/hashset.h
Me-k-01: Same as previous comment, we could just prevent it's use for now and avoid further Regular / CUDA code splitting.

Ponca/src/SpatialPartitioning/KdTree/Query/kdTreeKNearestQueries.h
Me-k-01: My bad, I must have though Base was used as self in the Ponca API, I'll change it back to Self, and also update `LimitedPriorityQueue` to alias itself to Self instead of Base, as the same mistake was made there too.

Ponca/src/SpatialPartitioning/KdTree/kdTree.h
Me-k-01: Good catch, it's not true anymore, I'll update the warnings in future commits.

Ponca/src/SpatialPartitioning/KdTree/kdTreeTraits.h
Me-k-01: This value is badly named, because even if in the Stack we use it to initialize the size of the array that will contain all of the K-nearest neighbors, it's also used as an indicator of what's the highest possible indice in the point cloud (corresponding the the max point cloud size) when we use the BitSet.

The current BitSet logic requires that we know in advance how big the point cloud is. Because we need to allocate a bit for each possible indice (this is all commented out in the BitSet class and KnnGraphRangeNeighborQuery class btw). The indices that we store in the BitSet must always be bellow this number.

For the HashSet we aren't limited like that. Maybe we should default to the HashSet as it doesn't have to know in advance what's the total number of point in the point cloud?

Me-k-01: I'll update to default to the HashSet for now, and reduce this value as suggested, until we decide if we want to keep the BitSet logic or not

tests/common/kdtree_utils.h
Me-k-01: I don't see any missing line on my end???

Ponca/src/Common/Containers/bitset.h
Me-k-01: In latest commits I've split the code for now to be using std::fill when not in CUDA, but I honestly think it's bad practice.

tests/common/kdtree_utils.h
Me-k-01: Ah, there was two end of line at the end of the file. One seems to have been automatically removed by the clang format tool.

Ponca/src/Common/Macro.h
```diff
@@ -28,17 +30,27 @@
...
+        PONCA_MACRO_START                                                \
+        fprintf(stderr, "%s:%i: [Error] %s\n", __FILE__, __LINE__, MSG); \
+        fflush(stderr);                                                  \
+        PONCA_MACRO_END
+#    define PONCA_PRINT_WARNING(MSG)                                       \
+        PONCA_MACRO_START                                                  \
+        fprintf(stderr, "%s:%i: [Warning] %s\n", __FILE__, __LINE__, MSG); \
+        fflush(stderr);                                                    \
+        PONCA_MACRO_END
+#endif
```
Me-k-01, line 53: To be removed from this PR once #311 is merged

examples/cuda/ponca_fit_knngraph.cu
Me-k-01: That was to test that the result of the RangeQuery was the same when using the KnnGraph compared to the KdTree, independently of the KnnGraph culling. We can use a lower value now, I suppose point.size() / 4 like the tests should be a good example?

Me-k-01:
After decreasing the k nearest neighbor value in the CI test, it seems coverage doesn't reach the case where our limited priority queue reach it's max size. Further testing is required to know if that's a bad thing or not.

Ponca/src/Common/Containers/limitedPriorityQueue.h
Me-k-01: Done

Me-k-01:
I've added a new test to the CI for the BitSet and HashSet, but we still need to add a better doc about the use cases of the HashSet inside the `KnnGraphRangeNeighborsQuery`. I've also made the hashing function replaceable through a lambda function.
I've tweaked the push method of the `LimitedPriorityQueue` for easier maintainability, and added it to the new `common_container` test on the CI.

This PR is starting to get big so I think the remainings CUDA warnings will have to wait for another PR...

**EDIT:** Updating the kernel resolved the weird warnings on my end.

## Remaining tasks :
- [x] Investigate the coverage results of the knearest neighbor test
- [x] Fix CLang

Me-k-01:
Ready for review @nmellado (still working on finishing the doc for the RangeQuery)

examples/cuda/ponca_fit_knngraph.cu
```diff
@@ -0,0 +1,168 @@
...
+struct KnnGraphKNearestFunctor {
+    static __device__ inline auto query(
+        KnnGraphGPU<DataPoint>& d_knngraph,
+        int i, typename DataPoint::Scalar analysisScale
+    ) -> Ponca::KnnGraphKNearestQuery<Ponca::KnnGraphPointerTraits<DataPoint>> {
+        //! [Use KnnGraph.kNearestNeighbors on the GPU]
+        return d_knngraph.kNearestNeighbors(i);
+        //! [Use KnnGraph.kNearestNeighbors on the GPU]
+    }
+};
```
nmellado: Move to cuda_utils ? as it is used in both example (except if I missed a detail).

Ponca/src/Common/Containers/hashset.h
```diff
@@ -0,0 +1,137 @@
...
+     * on the negative value you need to store to avoid this issue, with the following rule : illegal_value =
+     * EMPTY-OFFSET (e.g. set OFFSET to 2 to allow to store -1, but make -2 illegal to store).
+     *
+     * For the search, the best case complexity is O(1), and the worst case complexity is O(n).
+     * The complexity is entirely dependent on the hashing function, and the given values :
+     * If our hashing function produce sparser result for our set of indices, it will reduce the complexity of the
+     * searches, and therefore improve the performance of our algorithm.
+     *
+     * \see HashSet::hash for the hashing function
+     *
```
nmellado, line 47: Maybe links to other `Bitset`, e.g.:
```
\see BitSet For alternative data structure with compatible API
```
(and the other way around in Bitset.

Ponca/src/SpatialPartitioning/KdTree/Iterator/kdTreeKNearestIterator.h
```diff
@@ -22,7 +24,7 @@ namespace Ponca
      *
      *  \see KdTreeKNearestQueryBase
      */
```
nmellado, line 28: Add doc for `\tparam MAX_KNN_SIZE`

Me-k-01 (commented):
Non usefull comments were removed.
This PR is ready for review @nmellado

nmellado (dismissed):

PR #302, opened 2026-04-01
[main] CPM support

BDoignies:
Hi, 

This PR adds documentation about CPM and its ponca support. 

I also added a github action check, but this is linked to: [https://github.com/poncateam/ci-submodule-helper/pull/4](https://github.com/poncateam/ci-submodule-helper/pull/4). Thus, it will fails the action until the corresponding PR is merged.


nmellado:
I updated the helper repository. It fails in some situations.

BDoignies:
Yes, I forgot about Eigen...

BDoignies:
@nmellado The last commit should ensure that Eigen is always available for transitive linking. This is important to me because otherwise, using ponca within cmake is inconsistent (between just linking and having to include Eigen directories).  

This is tricky, but it seems to work. Another option would be to recreate Eigen's target, but managing the install/export would be problematic. 

Is it important to provide the Eigen version and directory? For now, they are only used within Ponca for messages. Providing the version is a bit tricky. The line we should write can be found at [Line 103 of Eigen's CMakeLists](https://gitlab.com/libeigen/eigen/-/blob/master/CMakeLists.txt?ref_type=heads#L103)

Me-k-01 (dismissed):

nmellado (dismissed):

nmellado (approved):

PR #304, opened 2026-04-14
Compile time

BDoignies:
Hi, 

This PR proposes to reduce compile time with a combination of precompiled headers and explicit instantitation.

Notably, this PR introduces:
- New CMake target Precompiled (alias Ponca::Precompiled)
- New CMake options to control what is pre-instantiated

What's missing:
- Examples and test should likely use this library instead
- Documentation
- A better list of basket. For now I instantiate those used in ponca_basic_cpu to see that everything works fine. 

Currently, the 'library' is compiled as a shared dynamic library, but I did not think much about what's the best. We may alway add an option to select [how this is linked](https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html). Some build may require a __dllspec(dllexport) on windows or a Position Independant Code build. Also, we should look [into this page](
https://gcc.gnu.org/onlinedocs/gnat_ugn/Compilation-options.html) if we decide to go for a static library. 

As discussed AFK with @nmellado, I did not use heavy Expand/[X-macros](https://en.wikipedia.org/wiki/X_macro) tricks. Instead, I decided to go with a conceptually simpler, yet equivalent method. Whitin the Ponca/precompiled/instantiate folder, files defines only what they need to, and includes other files. There is a bit of code duplication but not much in the end. This build a cartesian product of all options and is very modular. It also allows for *easy* filtering of cases that should not exist (GLS requires 3D and Normals for instance). 
There is still some traces of Xmacros for now (see Ponca/precompiled/instantiate/baskets.h) because we will need the whole list in the future for the factory (see paragraph bellow) as well as the python binding (if it ever exist). 

Speaking about the factory (that should be part of another PR I think), @nmellado @Me-k-01 I'd like your opinion. How should the user specify what he cares about? For example, whether he wants a Fit that uses a SmoothWeightKernel or a GaussianKernel? I have *two* options in mind:

1. We name instances as follow: GLSSphere_3fPPNSW for a 'GLS Sphere fit', in 3D, with Float, PointNormalPosition and a SmoothWeight kernel. ANy other format is fine, we just need to pick one.
2. We build subfactories. With the same example, something like: Factory<float, 3, PointPositionNormal<float, 3>, DistWeightFunc<...>> that only contains the relevant options. 

We can also think about a mix of the two above. For example having a method: GetNames<float, 3>() that returns a list of all compute object that will use 3D float points (or another object for further filtering).


cmake/PoncaConfigurePrecompiled.cmake
```diff
@@ -0,0 +1,91 @@
...
+install(EXPORT PrecompiledTargets
+  FILE PoncaTargets-Precompiled.cmake
+  NAMESPACE Ponca::
+  DESTINATION lib/cmake/Ponca
+)
+
+export(EXPORT CommonTargets
+  FILE ${CMAKE_CURRENT_BINARY_DIR}/PoncaTargets-Precompiled.cmake
+  NAMESPACE Ponca::
+)
```
Me-k-01:
```suggestion
)

```

BDoignies:
@nmellado I rebased the branch. 

I also made sure examples use the Ponca::Precompiled target but I did not add the relevant baskets for now. 

For tests, this is more complicated. Most of the test are really small. Hence, the cost of the precompiled header overweigth the gain (cmake will create one for every test but this compiles everything...). I am not exactly sure why this is so long. I'll investigate this but not sure parsing the whole library is meaningfull for a test about filters ^^'. An  option would be to disable precompiled_header.  

I think bencharking this will be necessary.

cmake/PoncaConfigureInstantiate.cmake
```diff
@@ -0,0 +1,103 @@
...
+        )
+    foreach(param IN LISTS ponca_Instantiate_OPTIONS)
+        option(${param} "" OFF)
+        if (${param})
+            target_compile_definitions(Instantiate PUBLIC _${param})
+        endif()
+    endforeach()
+endif()
+
+
```
nmellado:
```suggestion
# Required as the target does not have source files, which are used by cmake to infer the linker type.
```

nmellado (approved):

PR #306, opened 2026-04-15
[spatialPartitioning] Set kdtree buffer size after container conversion

### LINKED ISSUE #305: Regression with point container type applications
The following code from Poncascope doesn't work anymore with latest release
```
template<typename KdTreeType>
void buildKdTree(const Eigen::MatrixXd& cloudV, const Eigen::MatrixXd& cloudN, KdTreeType& tree){
    std::vector<int> ids(cloudV.rows());
    std::iota(ids.begin(), ids.end(), 0);

    using VN = std::pair<const Eigen::MatrixXd&, const Eigen::MatrixXd&>;

    // Build KdTree: do not copy coordinate but rather store Eigen::Block
    tree.buildWithSampling(VN(cloudV, cloudN),
                           ids,
                           [](VN bufs, typename KdTreeType::PointContainer &out) {
                               int s = bufs.first.rows();
                               out.reserve(s);
                               for (int i = 0; i != s; ++i)
                                   out.push_back(typename KdTreeType::DataPoint(bufs.first.row(i).transpose(),
                                                                                bufs.second.row(i).transpose()));
                           });
}
```

This is caused by the change made in `KdTreeBase<Traits>::buildWithSampling`, where we now call `points.size()` to initialize the `Base::m_bufs.points_size` value, which isn't available with `std::pair`. 

# Suggested change

We should fetch the size value after converting the container to its internal type (of `std::vector`). We won't be calling this function on internal pointer type anyway, since `KdTreeBase` methods are only defined for containers that are STL-like.

Me-k-01:
- Fix regression #305 by setting the buffer size after the container conversion
- Avoid Forwarding reference warning

### DIFF
--- a/Ponca/src/SpatialPartitioning/KdTree/kdTree.hpp
+++ b/Ponca/src/SpatialPartitioning/KdTree/kdTree.hpp
@@ -123,10 +123,10 @@ PONCA_MULTIARCH_HOST inline void KdTreeBase<Traits>::buildWithSampling(PointUser
 
     // Move, copy or convert input samples
     c(std::forward<PointUserContainer>(points), Base::m_bufs.points);
-    Base::m_bufs.points_size = points.size();
+    Base::m_bufs.points_size = Base::m_bufs.points.size();
 
-    Base::m_bufs.indices_size = sampling.size();
-    Base::m_bufs.indices      = std::move(sampling);
+    Base::m_bufs.indices      = std::forward<IndexUserContainer>(sampling);
+    Base::m_bufs.indices_size = Base::m_bufs.indices.size();
 
     Base::m_bufs.nodes.reserve(4 * Base::pointCount() / Base::m_min_cell_size);
     Base::m_bufs.nodes.emplace_back();
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -6,9 +6,10 @@ Current head (v.2.0.alpha3 RC)
 This release will introduce a refactoring of the requirement library using cpp20::concepts.
 
 - Bug-fixes and code improvements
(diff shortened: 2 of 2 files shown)


Ponca/src/SpatialPartitioning/KdTree/kdTree.hpp
```diff
@@ -123,10 +123,10 @@ PONCA_MULTIARCH_HOST inline void KdTreeBase<Traits>::buildWithSampling(PointUser
 
     // Move, copy or convert input samples
     c(std::forward<PointUserContainer>(points), Base::m_bufs.points);
-    Base::m_bufs.points_size = points.size();
+    Base::m_bufs.points_size = Base::m_bufs.points.size();
 
-    Base::m_bufs.indices_size = sampling.size();
-    Base::m_bufs.indices      = std::move(sampling);
+    Base::m_bufs.indices      = std::forward<IndexUserContainer>(sampling);
```
nmellado, line 128: This PR adds two changes : 
- calling `.size()` on the STL container (instead of the custom one)
- replacing a `move` by a forward.

I got the rational for the first change (#305), but not for the second one. Could you please elaborate ?

Me-k-01: Using Clion I was getting the following warning with `std::move` : 

```
Clang-Tidy: Forwarding reference passed to std::move(), which may unexpectedly cause lvalues to be moved; use std::forward() instead 
```

If it's irrelevant I can change it back to `std::move`

nmellado: it's ok, just wanted to have the rationale.

nmellado (approved):

PR #311, opened 2026-04-21
[common] Adapt `PONCA_ASSERT` macro to CUDA

### LINKED ISSUE #310: The `PONCA_ASSERT` macro isn't compatible with CUDA
Using this macro in the CUDA kernel provokes the following error and warnings : 

```sh
~/CLionProjects/poncascope/external/ponca/Ponca/src/SpatialPartitioning/KnnGraph/Query/knnGraphRangeQuery.h(100): warning #20011-D: calling a __host__ function("fprintf") from a __host__ __device__ function("Ponca::KnnGraphRangeQuery< ::Ponca::KnnGraphPointerTraits< ::Ponca::PointPositionNormal<float, (int)3> > > ::advance") is not allowed

~/CLionProjects/poncascope/external/ponca/Ponca/src/SpatialPartitioning/KnnGraph/Query/knnGraphRangeQuery.h(100): error: identifier "stderr" is undefined in device code

~/CLionProjects/poncascope/external/ponca/Ponca/src/SpatialPartitioning/KnnGraph/Query/knnGraphRangeQuery.h(100): warning #20011-D: calling a __host__ function("fflush") from a __host__ __device__ function("Ponca::KnnGraphRangeQuery< ::Ponca::KnnGraphPointerTraits< ::Ponca::PointPositionNormal<float, (int)3> > > ::advance") is not allowed

~/CLionProjects/poncascope/external/ponca/Ponca/src/SpatialPartitioning/KnnGraph/Query/knnGraphRangeQuery.h(100): warning #20011-D: calling a __host__ function("__builtin_trap") from a __host__ __device__ function("Ponca::KnnGraphRangeQuery< ::Ponca::KnnGraphPointerTraits< ::Ponca::PointPositionNormal<float, (int)3> > > ::advance") is not allowed
```

Using regular assert is allowed in CUDA, maybe we should simply switch to that?

Me-k-01:
Fix #310 by taking into account `#ifdef __CUDA_ARCH__`
The program is simply aborted with `asm("trap;")` if we are on the device CUDA. We now can ensure that the program is stopped if we encounter an error on either the host or the device, unlike `assert` which doesn't terminate abruptly the CUDA kernel.

### DIFF
--- a/Ponca/src/Fitting/cncFormulaEigen.h
+++ b/Ponca/src/Fitting/cncFormulaEigen.h
@@ -61,6 +61,7 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 #pragma once
 
+#include <Ponca/src/Common/Assert.h>
 #include <Eigen/Dense>
 
 namespace Ponca::internal
@@ -310,8 +311,8 @@ namespace Ponca::internal
 
             // SelfAdjointEigenSolver returns sorted eigenvalues, no
             // need to reorder the eigenvectors.
-            assert(eigensolver.eigenvalues()(0) <= eigensolver.eigenvalues()(1));
-            assert(eigensolver.eigenvalues()(1) <= eigensolver.eigenvalues()(2));
+            PONCA_ASSERT(eigensolver.eigenvalues()(0) <= eigensolver.eigenvalues()(1));
+            PONCA_ASSERT(eigensolver.eigenvalues()(1) <= eigensolver.eigenvalues()(2));
             VectorType v1 = eigensolver.eigenvectors().col(1);
             VectorType v2 = eigensolver.eigenvectors().col(0);
             return std::pair<VectorType, VectorType>(v1, v2);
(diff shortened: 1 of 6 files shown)


Me-k-01:
Ready for review @nmellado

Me-k-01:
Ready for review @nmellado.
Unsure why codecov isn't happy, since it shows a -40.00% decrease in coverage for `Ponca/src/SpatialPartitioning/KnnGraph/Iterator/knnGraphRangeIterator.h` despite that I didn't changed this file.

Me-k-01:
It seems i'm not getting an error when compiling with CUDA and trying the PONCA_ASSERT inside a host function (`asm("trap;")` seems to only work in the device). 
I'll move this PR as a draft for now until this gets fixed

Me-k-01:
The bug is fixed, this PR is ready for review @nmellado

Ponca/src/Fitting/cncFormulaEigen.h
```diff
@@ -61,6 +61,7 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 #pragma once
 
+#include <Ponca/src/Common/Assert.h>
```
nmellado: Please check, but if I remember well we use relative paths inside the library,

Me-k-01: There is no mention of this currently in the guidelines, but now I'm looking at the other headers, yes it does seem that we are overall using relative path.
I'll push a commit to change this to relative path.

Me-k-01:
Ready for review @nmellado

Me-k-01:
This PR is still pending for a review @nmellado

nmellado (approved):

PR #316, opened 2026-04-24
[fitting] Fix CNC not using the DistWeightFunc

### LINKED ISSUE #313: [fitting] CNC is currently not filtering the neighborhood
CNC is currently using : 
```cpp
        using NeighborFilter = NeighborFilterStoreNormal<DataPoint, NoWeightFunc<DataPoint>>;
```
But we need to filter depending on the neighborhood of the points with `w(points[index]).first == Scalar(0.)`. We should therefore use a DistWeightFunc instead. Previous tests on the CI weren't catching this issue because we were filtering using a kdtree.

Me-k-01:
Fix #313

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -11,6 +11,7 @@ This release will introduce a refactoring of the requirement library using cpp20
 
 - Bug-fixes and code improvements
     - [fitting] Fix warnings introduced when bumping to cxx20 (#303)
+    - [fitting] Fix CNC not using the DistWeightFunc (#316)
     - [spatialPartitioning] Fix regression introduced by the CUDA kdtree update (#306)
     - [common] Fix PONCA_ASSERT macro on MSVC and clang (#309)
     - [common] Adapt PONCA_ASSERT macro to CUDA (#311)
--- a/Ponca/src/Fitting/cnc.h
+++ b/Ponca/src/Fitting/cnc.h
@@ -112,13 +112,14 @@ namespace Ponca
         };
 
     public:
-        using DataPoint      = P;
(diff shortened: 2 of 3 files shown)


Me-k-01:
@nmellado this PR is ready for review.

The clang formatting tool forced me to change the other lines to pass the test.

CHANGELOG
```diff
@@ -13,6 +13,7 @@ This release will introduce a refactoring of the requirement library using cpp20
     - [fitting] Fix warnings introduced when bumping to cxx20 (#303)
     - [spatialPartitioning] Fix regression introduced by the CUDA kdtree update (#306)
     - [common] Fix PONCA_ASSERT macro on MSVC and clang (#309)
+    - [fitting] Fix CNC not using the DistWeightFunc (#316)
```
nmellado:
```suggestion
    - [fitting] Fix warnings introduced when bumping to cxx20 (#303)
    - [fitting] Fix CNC not using the DistWeightFunc (#316)
    - [spatialPartitioning] Fix regression introduced by the CUDA kdtree update (#306)
    - [common] Fix PONCA_ASSERT macro on MSVC and clang (#309)
```

(don't use the suggestion directly, it seems to be broken, just please group fitting entries).

Me-k-01: Like this new commit?

nmellado: No, keep the two different lines, put all the fitting lines, then spatialPartioning, then the common, etc..

Me-k-01:
I may have found a pretty massive bug with the way we construct the NeighborFilters that affects every fits... I'll open an issue post for it

Me-k-01:
Ready for review @nmellado

PR #321, opened 2026-05-04
[spatialPartitioning] Fix points reference change in KnnGraph

### LINKED ISSUE #319: Bug With KnnGraph construction
In the code recently pushed in the `master` branch (#263), the KnnGraph moves the KdTree point container at construction time:

https://github.com/poncateam/ponca/blob/911e152b8d5ac5c934a260b3832f7f62800b65b9/Ponca/src/SpatialPartitioning/KnnGraph/knnGraph.h#L202-L208

The method `KdTree::Points()` returns a reference to this inner container:
https://github.com/poncateam/ponca/blob/911e152b8d5ac5c934a260b3832f7f62800b65b9/Ponca/src/SpatialPartitioning/KdTree/kdTree.h#L223-L227

This seems to be a bad behavior as it transfers the ownership of this container from the KdTree to the KnnGraph. 
This is definitely a problem when buffers are `std::vector`.
In Poncascope, it leads to unwanted copies of the point container (does not compile with the `master` branch anymore.

In previous versions of the code, we were using references, which I believe make more sense
https://github.com/poncateam/ponca/blob/961a4c6d64c1f300c899ec6da23c269ee5d6623f/Ponca/src/SpatialPartitioning/KnnGraph/knnGraph.h#L177

This needs to be fixed ASAP.

### LINKED ISSUE #325: [spatialParitioning] Bug with `KnnGraphRangeQuery` on CUDA
Using `KnnGraphRangeQuery::advance` provokes  illegal memory access on the current master.

```
Using range neighbors query :
Number of nodes in the KdTree : 15
CUDA error: an illegal memory access was encountered at /home/fauberva/CLionProjects/poncascope/external/ponca/examples/cuda/ponca_fit_knngraph.cu:101

Process finished with exit code 134 (interrupted by signal 6:SIGABRT)
```

Me-k-01:
- [x] Make the KnnGraph reference the kdtree PointsContainer
- [x] Fix CI KnnGraphPointersTraits test for reference change
- [x] Ensure that master of PoncaScope compiles with this proposal
- [x] Fix CUDA KnnGraph::Buffers transfer to device -> Opening new issue #325 for the remaining illegal memory access error as it is unrelated to the changes made by this PR


nmellado:
See call to copy constructor when compiling poncascope with the current `master`

```
../poncascope/external/ponca/Ponca/src/SpatialPartitioning/KnnGraph/knnGraph.h:208:38: note: in instantiation of member function 'std::vector<BlockPointAdapter<double>>::operator=' requested here
  208 |             Base::m_bufs.points      = std::move(_kdtree.points());
      |                                      ^
../poncascope/src/datastructures.cpp:46:32: note: in instantiation of function template specialization 'Ponca::KnnGraphBase<Ponca::KnnGraphDefaultTraits<BlockPointAdapter<double>>>::KnnGraphBase<Ponca::KdTreeDefaultTraits<BlockPointAdapter<double>>>' requested here
   46 |         context.knnGraph = new Context::KnnGraph(context.tree, context.kNNGraphK);
      |                                ^
../poncascope/src/./poncaAdapters.hpp:48:18: note: copy assignment operator of 'BlockPointAdapter<double>' is implicitly deleted because field 'm_pos' has no copy assignment operator
   48 |     InternalType m_pos;
      |                  ^
6 warnings and 1 error generated.
```

nmellado:
@Me-k-01 can you please check if this proposal works well with cuda ?

Me-k-01:
This proposal doesn't and cannot work with CUDA : As long as we have reference, we won't be able to initialize an empty knngraph on the GPU. 

We could try replacing the reference with a pointer instead (with default empty constructor set to nullptr). The following commits will attempt to reconcile Poncascope and the CUDA implementation using this proposed pointer alternative.

Me-k-01:
For the moment, CUDA compiles but gives an unspecified error during the launch of the kernel
```cpp
./ponca/cmake-build-debug/examples/cuda/cuda_fit_plane_knngraph
Example plane fitting using KnnGraph on CUDA...
Number of nodes in the KdTree : 43
CUDA error: unspecified launch failure at /home/fauberva/CLionProjects/poncascope/external/ponca/examples/cuda/ponca_fit_knngraph.cu:100

Process finished with exit code 134 (interrupted by signal 6:SIGABRT
```

Ponca/src/SpatialPartitioning/KnnGraph/Query/knnGraphRangeQuery.h
```diff
@@ -136,8 +136,50 @@ namespace Ponca
...
+        PONCA_MULTIARCH inline void advance(Iterator& iterator)
+        {
+            const auto& points = m_graph->points();
+            const auto& point  = points[QueryType::input()].pos();
+
+            if (iterator == end())
+                return;
+
+            // Search the next neighbor in range
+            while (!m_stack.empty())
```
nmellado, line 125: The problem here is that you can `pop()` an element x1, access its neighbor x2, and add x1 again as it is a neighbor of x2. I believe that the call to `pop` must be done after the fore loop, e.g., around line 177.

examples/cuda/cuda_utils.cu
nmellado, line 209: Is `cudaFree` also called on the GPU buffer structure ? (which is allocated with CudaMalloc ?

examples/cuda/cuda_utils.cu
```diff
@@ -182,14 +183,17 @@ void deepCopyKdTreeBuffersToDevice(const KdTreeBuffers& hostBuffers, // Input
...
+void deepCopyKnnGraphBuffersToDevice(
+    const KnnGraphBuffers& hostBuffers, // Input
+    StaticKnnGraphBuffers& hostBuffersHoldingDevicePointers,
+    StaticKnnGraphBuffers* const deviceBuffers // Outputs
 )
 {
-    deepCopyBuffersToDevice<Traits>(
-        hostBuffers, [&]() { hostBuffersHoldingDevicePointers.k = hostBuffers.k; }, hostBuffersHoldingDevicePointers,
-        deviceBuffers);
+    CUDA_CHECK(
```
nmellado: Why don't we have a `CudaMalloc` here ? (we do have it in the KdTree version `deepCopyKdTreeBuffersToDevice`)

Ponca/src/Common/Containers/hashset.h
```diff
@@ -71,19 +71,16 @@ namespace Ponca
...
         constexpr PONCA_MULTIARCH HashSet() : m_data()
         {
-            // Skip this initialization step if EMPTY is set to 0
-            if constexpr (EMPTY != T(0))
-            {
-                Ponca::internal::fill(m_data, m_data + N, EMPTY);
-            }
+            // Skip this array initialization step if the EMPTY flag is set to 0
+            if constexpr (EMPTY() != T(0))
+                Ponca::internal::fill(m_data, m_data + N, EMPTY());
```
nmellado: I don't understand this as EMPTY() always equals 0, and cannot be changed. What the matter ?

nmellado (commented):
I didn't test the code, but I might have spotted some issues, including with the memory management (only by reading the code, so can be wrong), that could explain the runtime issues you are experiencing.

Ponca/src/Common/Containers/hashset.h
Me-k-01: I wanted to have some modularity on this structure, that is why I added this check on the constructor, to properly fill the array if the EMPTY flag is not zero. This way we can eventually add EMPTY and OFFSET as a template argument to the structure, and make them available to change. But it might be a little bit too in depth to the user to have access to this?

Since I added the OFFSET flag, the EMPTY one has become a bit useless. We can remove this and simply use zero everywhere instead. 

It is a bit bothersome in CUDA to have this `if constexpr`, because the static constexpr value isn't defined on the device at compile time when it checks for this `if constexpr` expression, even when I add a device flag. I couldn't get it to work unless I called it through a static constexpr `EMPTY()` function like this, for some strange reason. Now that I'm thinking of this again, defining EMPTY though template arguments might solve this one issue.

Ponca/src/SpatialPartitioning/KnnGraph/Query/knnGraphRangeQuery.h
Me-k-01: That is not supposed to happen because we use `m_flag.insert(idx_nei)` to remember that we should not add the same point again to the stack. This is the same as we've done with the original recursive algorithm. The problem shouldn't lie here since we also get the error when using the original recursive algorithm.

examples/cuda/cuda_utils.cu
Me-k-01: Oof that `cudaMemcpy(deviceBuffers..` line was misplaced here. This code shouldn't be changed compared to master. Happened when I was debugging the buffers. This was unintended.

examples/cuda/cuda_utils.cu
Me-k-01: Yes, it is called at line 112

Ponca/src/Common/Containers/hashset.h
Me-k-01: For a cleaner and simpler structure, I've removed EMPTY alltogether and made OFFSET a template parameter of the class. Can you review the change?

Ponca/src/Common/Containers/hashset.hpp
```diff
@@ -40,29 +66,31 @@ namespace Ponca
...
+        // A - The set is full (The last search index is -1 if it didn't find an available address in the array)
+        if (availableIdx == -1)
+        {
+            return std::make_pair(end(), false);
+        }
         // B - The set isn't full and the value can be inserted. Therefore, the last search index is the next available
         // address in the array
         m_data[availableIdx] = _value + OFFSET;
-        return true;
+        return std::make_pair(m_data.begin() + availableIdx, true);
```
Me-k-01, line 87: In this new version we now return a pair of iterator / boolean to allow to check if the set is full : returned_pair = {end(), false}

Me-k-01 (commented):
Ready for review @nmellado

Ponca/src/Common/Containers/hashset.hpp
Me-k-01: This is done this way to be compatible with the std::set insert method

Me-k-01: But currently the iterator isn't a real set iterator that can iterate over only the element stored in the set. It is just an array::iterator

nmellado (commented):
I do not understand how we can change the types used for IndexSet and Stack. The API is strange. Fist, there are two classes with the same name:

- [KnnGraph/Query/knnGraphRangeQuery.h#L49-L50](https://github.com/Me-k-01/ponca/blob/04c343644dce3a490807fb5012b7f4730ca58dea/Ponca/src/SpatialPartitioning/KnnGraph/Query/knnGraphRangeQuery.h#L49-L50):
```cpp
    template <typename Traits, typename IndexSet, typename Stack>
    class KnnGraphRangeQuery : public RangeIndexQuery<typename Traits::IndexType, typename Traits::DataPoint::Scalar>
```
and 
- [KnnGraph/Iterator/knnGraphRangeIterator.h#L38-L44](https://github.com/Me-k-01/ponca/blob/04c343644dce3a490807fb5012b7f4730ca58dea/Ponca/src/SpatialPartitioning/KnnGraph/Iterator/knnGraphRangeIterator.h#L38-L44)
```cpp
    template <typename Traits, typename Set = std::set<int>,
              // typename Set = Ponca::BitSet<100000>,
              // typename Set = Ponca::HashSet<Traits::MAX_RANGE_EXPLORATION_AMOUNT>,
              typename Stack = std::stack<int>
              // typename Stack = Ponca::Stack<int, Traits::MAX_RANGE_EXPLORATION_AMOUNT>
              >
    class KnnGraphRangeQuery;
```

In [KnnGraph/knnGraph.h#L61](https://github.com/Me-k-01/ponca/blob/04c343644dce3a490807fb5012b7f4730ca58dea/Ponca/src/SpatialPartitioning/KnnGraph/knnGraph.h#L61), you use the second definition
```cpp
        using RangeIndexQuery    = KnnGraphRangeQuery<Traits>;
```
that you construct and return in 
```cpp
PONCA_MULTIARCH inline RangeIndexQuery rangeNeighborsIndexQuery() const { return RangeIndexQuery(this, 0, 0); }
```

This is confusing as you use this version in Cuda as well, while we would use a version without the set, right ? See 
[examples/cuda/cuda_utils.cu#L68-L79](https://github.com/Me-k-01/ponca/blob/04c343644dce3a490807fb5012b7f4730ca58dea/examples/cuda/cuda_utils.cu#L68-L79)
```cpp
template <typename DataPoint>
struct KnnGraphRangeFunctor
{
    static __device__ inline auto query(KnnGraphGPU<DataPoint>& d_knngraph, int i,
                                        typename DataPoint::Scalar analysisScale)
        -> Ponca::KnnGraphRangeQuery<Ponca::KnnGraphPointerTraits<DataPoint>>
    {
        //! [Use KnnGraph.rangeNeighbors on the GPU]
        return d_knngraph.rangeNeighbors(i, analysisScale);
        //! [Use KnnGraph.rangeNeighbors on the GPU]
    }
};
```

Me-k-01:
Yes, the API still needed some modifications to allow the user to change the Stack and Set depending on the use case. 

This last commit reworks the KnnGraphRangeQuery API to use the Stack, Set and const_iterator provided by the Traits type. I've set the KnnGraphDefaultTraits to use the original std version of the Stack and the Set. I've also changed the KnnGraphPointerTraits to use Ponca::HashSet and Ponca::Stack. This Traits type is used by the CUDA examples and works just fine, it is also tested in the CI.

It seems from the error logs that the changes you made to the `Iterator(m_graph->getIndexPtr())` breaks the compatibility with the previously working call to `std::vector<IndexType>::const_iterator` constructor on MacOs
```sh
/Users/runner/work/ponca/ponca/src/Ponca/src/SpatialPartitioning/KnnGraph/Query/knnGraphKNearestQuery.h:62:20: error: calling a private constructor of class 'std::__wrap_iter<const int *>'
   62 |             return Iterator(m_graph->getIndexPtr()) + QueryType::input() * m_graph->k();
      |                    ^
```
https://github.com/poncateam/ponca/actions/runs/25657598351/job/75309739091?pr=321

This is wasn't the case before the change with `getIndexPtr`. This also doesn't happen when we use `typename std::array<const IndexType, K_MAX_NN>::const_iterator`. This is a bit strange, could you investigate this issue @nmellado ?

Ponca/src/SpatialPartitioning/KnnGraph/Query/knnGraphKNearestQuery.h
```diff
@@ -40,7 +40,7 @@ namespace Ponca
 #endif
     {
     public:
-        using Iterator = typename Traits::KnnGraphRangeIterator;
+        using Iterator = KnnGraphKNearestIterator<const typename Traits::IndexType*, typename Traits::IndexType>;
```
Me-k-01, line 43: I've added back KnnGraphKNearestIterator class to solve the `std::vector<int>::const_iterator` constructor being set to private for pointers on MacOS and Windows. This last version works for Poncascope and CUDA. 

Let me know what you think @nmellado

PR #327, opened 2026-05-12
[SpatialPartitioning] Add generic neighbor graphs

nmellado:
Extends the KnnGraph API to be able to use neighbor-graphs with variable neighborhood sizes.

### Todo

- [x] Create base classe for generic neighbor graphs
- [x] Move queries to base classes
- [x] Create class with variable neighborhood sizes
- [x] Add non-uniform neighborhood graph creation from Kdtree range query
- [x] Add tests
- [x] Update documentation
- [x] Improve class names (RangeNeighborGraph)
- [x] Visualize graphs in Poncascope, see https://github.com/poncateam/poncascope/pull/33

### Visualizations

- KnnGraph 
Capture d’écran 2026-05-15 à 15 01 14

- Range NeighborGraph (r=0.03)
Capture d’écran 2026-05-15 à 15 01 49

- Range NeighborGraph (r=0.04)
Capture d’écran 2026-05-15 à 15 01 40

- Range NeighborGraph (r=0.1, very large)
Capture d’écran 2026-05-15 à 15 00 58


Ponca/src/SpatialPartitioning/NeighborGraph/Iterator/neighborGraphOneConnectedIterator.h
```diff
@@ -7,48 +7,51 @@
...
      *
      *  \warning This iterator object should never be duplicated, as it is a proxy that holds a reference to the actual
-     * data : The copy of this iterator would still point to the same KnnGraph reference. So, if the increment operator
-     * is used on the iterator, the duplicate will also have its state updated. If we then call the increment operator
-     * on the duplicate, the result will be an incorrect value.
-     *
-     *  \see KnnGraphKNearestQueryBase
+     * data : The copy of this iterator would still point to the same NeighborGraph reference. So, if the increment
+     * operator is used on the iterator, the duplicate will also have its state updated. If we then call the increment
+     * operator on the duplicate, the result will be an incorrect value.
```
BDoignies: I know that this PR does not introduce the sentence; but now that I see it: 

The warning is a bit more convoluted than it should be. I suggest: 
"data: modifying one will modify the others and can result in incorrect values." 

Which is more straightforward.

Ponca/src/SpatialPartitioning/NeighborGraph/abstractNeighborGraph.h
```diff
@@ -0,0 +1,159 @@
...
+        PONCA_MULTIARCH inline AbstractNeighborGraph(const Buffers& _bufs) : m_bufs(_bufs) {}
+
+        //! \brief Get the number of connection edges in the graph
+        PONCA_MULTIARCH [[nodiscard]] inline IndexType edgeCount() const { return (IndexType)m_bufs.indices_size; }
+        //! \brief Get the number of points
+        PONCA_MULTIARCH [[nodiscard]] inline IndexType pointCount() const { return (IndexType)m_bufs.points_size; }
+        //! \brief Get the internal point container
+        PONCA_MULTIARCH [[nodiscard]] inline PointContainer points() const { return m_bufs.points; };
+        //! \brief Get the internal index container
+        PONCA_MULTIARCH [[nodiscard]] inline IndexContainer edges() const { return m_bufs.indices; };
```
BDoignies, line 116: I would add const ref to those (manually, or using const IndexContainerRef or similar types), but it depends on whether we can guarantee PointContainer and IndexContainer are already references or not.

Ponca/src/SpatialPartitioning/NeighborGraph/knnGraph.h
```diff
@@ -0,0 +1,188 @@
...
+#pragma once
+
+#include "abstractNeighborGraph.h"
+#include "Query/neighborGraphOneConnectedQuery.h"
+#include "Query/neighborGraphRangeQuery.h"
+
+#include "../KdTree/kdTree.h"
+#include "../../Common/Assert.h"
+
+#define DEFAULT_K_IN_KNN_GRAPH 6
```
BDoignies: I would prefer a global constant within the ponca namespace instead.

Ponca/src/SpatialPartitioning/NeighborGraph/knnGraph.h
```diff
@@ -0,0 +1,188 @@
...
+    {
+#define WRITE_TRAITS                                                                                                    \
+    using Traits            = _Traits;                    /*!< Alias to the Traits type                          */     \
+    using DataPoint         = typename Traits::DataPoint; /*!< DataPoint given by user via Traits                */     \
+    using Scalar            = typename DataPoint::Scalar; /*!< Scalar given by user via DataPoint                */     \
+    using VectorType        = typename DataPoint::VectorType; /*!< VectorType given by user via DataPoint            */ \
+    using IndexType         = typename Traits::IndexType; /*!< Type used to index points into the PointContainer */     \
+    using PointContainer    = typename Traits::PointContainer; /*!< Container for DataPoint used inside the KdTree   */ \
+    using IndexContainer    = typename Traits::IndexContainer; /*!< Container for indices used inside the KdTree     */ \
+    using IndexContainerRef = typename Traits::IndexContainerRef; /*!< Ref type to index container */
```
BDoignies: This macros is used and defined multiples times in multiple files. Unless mistaken, this looks the same as "AbstractNeighborGraph" but without a macro. I suggest to move to a separate files either a defines.h within the KnnGraph folder or a common base file (such as neighborGraphTraits for examples).

BDoignies (dismissed):
LGTM. 
I have a few minor comments that you may choose to address or ignore. I’m still approving this since the comments concern pre-existing code issues and to not modify observable behavior.

Ponca/src/SpatialPartitioning/NeighborGraph/abstractNeighborGraph.h
nmellado: It might be tricky as they are either const references or pointers depending on the traits. If you have a solution don't hesitate to push it 👍

nmellado:
@BDoignies I updated the code according to your comments. The only remarks that remains is the question about the const qualifier for the buffers. Do you want to push it or shall we leave it at it is ?

PR #329, opened 2026-05-22
[fitting] Split BasketComputeObject and BasketDiffComputeObject

BDoignies:
Hi, 

As discussed with @nmellado, here is a proposal to split BasketComputeObject and BasketDiffComputeObject. The main idea is that currently, there is a false diamond inheritance that is expected to cause problem. Any modification to BasketComputeObject<Basket> would not be propagated to BasketComputeObject<BasketDiff> as they are different classes in different part of the hierarchy.  

For reference, the current class inheritance diagram is:

image

What we originally discussed with @nmellado was to make a new BasketDiffComputeObject that derives from BasketComputeObject. However, after more carefull consideration, I think that it is best to make it derive from ComputeObject insted:

1. Otherwise it would not solve the problem of modification propagation. Any BasketDiffComputeObject would have the same members as a BasketComputeObject because of inheritance: this duplicates members AND makes them wrongly available...
2. Semantically, they might be very different. I also find it more clear that the two don't interact and hence do not cause any "false diamond inheritence". 

With the proposal and according to the inheritance diagram below, the "BasketComputeObject" (and all its information) can be obtained by casting "BasketDiffComputeObject::this" to  the "BasketComputeObjectType*" type. 

image

Currently, a lot of code have been duplicated. I do not know whether this is the right time to make this PR. IMHO, it is best to simply add a comment introducing the problem and how to solve it (link to this PR?)

### DIFF
--- a/Ponca/src/Fitting/basket.h
+++ b/Ponca/src/Fitting/basket.h
@@ -130,7 +130,7 @@ namespace Ponca
      * \tparam Base Base class that provides, through the CRTP the init, startNewPass, addNeighbor and finalize methods
      */
     template <typename _Derived, typename _Base>
-    struct BasketComputeObject : public ComputeObject<_Derived>, public virtual _Base
+    struct BasketComputeObject : public ComputeObject<_Derived>, public _Base
     {
         using Base    = _Base;    /// <\brief Alias to the Base type
         using Derived = _Derived; /// \brief Alias to the Derived type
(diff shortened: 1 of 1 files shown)


nmellado:
Thanks for the proposal @BDoignies.
You are right, maybe we can post-pone this, I'll add an issue. So we can focus on the other big changes first.

BDoignies:
Closing. To be reopened later (issue #330)

PR #331, opened 2026-05-26
[doc] Update documentation for concepts

BDoignies:
Here is a small update of documentation for concepts. 

I decided not to go for a giant table within the documentation as it would not be much more informative than the Fitting/concepts.hpp file. The code is clear and straightforward and is the information. 

I also fixed a few typos and code / endcode markups.


doc/src/ponca_module_fitting.mdoc
```diff
@@ -273,27 +273,32 @@ namespace Ponca
...
-  
-  In order to ease tools combinations, each class declare a set of *required* and *provided* capabilities, which are detailed in the Section \ref concepts_provides.
+  In order to detect if the computational objects are correctly combined, Ponca relies on C++20 concepts. Most concepts provided for Basket extensions and tools can be found within the Fitting/concepts.hpp file. 
+  For instance, the GLSParam class is defined as:
+
+ ```cpp
+#define GLS_PARAM_REQUIREMENTS ProvidesAlgebraicSphere<T>
+template <class DataPoint, class _NFilter, typename T>
+    requires GLS_PARAM_REQUIREMENTS
+class GLSParam : public T;
```
nmellado: Can we use snippets instead ?

nmellado:
I like the idea of simplifying the documentation. This will be even more compact when we will have the factories.

doc/src/ponca_module_fitting.mdoc
BDoignies: That would mean two snippets for simply 3/4 lines. That seemed overkill to me. But we could.

nmellado: In my experience, code written explicitly in the documentation tends be deprecated very quickly. With snippets this cannot happen by design. So I would prefer, yes.

nmellado (dismissed):

nmellado (approved):

PR #333, opened 2026-05-27
Auto release on tag push

BDoignies:
This PR proposes to create github release for Ponca when a new tag is pushed (#44). It automatically parse the changelog in order to create the following release. At the time of writting this (the tag/release name is a dummy one for test purposes): 

image

[https://github.com/BDoignies/ponca/releases](https://github.com/BDoignies/ponca/releases)

Here are the rules:
* It looks for the text between the first set between "----". 
* It removes the first line (usually just the version name which is redondant)
* Prerelease and draft are infered from the tag name (alpha/beta within the name).

Assisted by: Claude (free version) for the awk command syntax.

### DIFF
--- a/.github/workflows/Release.yml
+++ b/.github/workflows/Release.yml
@@ -37,7 +37,7 @@ jobs:
         working-directory: build/install
         run: zip -r ../../ponca-${{ matrix.config.name }}.zip .
 
-      - name: Zipping install directory (Linux/Mac)
+      - name: Zipping install directory (MSVC)
         if: runner.os == 'Windows'
         working-directory: build/install
         run: Compress-Archive -Path * -DestinationPath ../../ponca-${{ matrix.config.name }}.zip
--- a/.github/workflows/Release.yml
+++ b/.github/workflows/Release.yml
@@ -1,0 +1,34 @@
+on:
+  push:
+    tags:
+      - 'v*'
+
+name: Create Release
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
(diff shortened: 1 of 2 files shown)


nmellado (approved):

nmellado (changes requested):
Please add artefact generation and publication

nmellado:
As you can see in #335, when we will publish the tag, the associated changelog is not between the first `------`. Could you update the script accordingly?

BDoignies:
What would work for you?

* If the tags match, I can try: find the line that have the tag then get the following lines until "---". 
* Get the second to third part ?

First one is more tricky, even a small mistake can make the release text fail...

BDoignies:
I first option: look for the version (without the 'v' and case insensitive) and return the block. If not found (we never know), a simple message is inserted instead "For information about changes, see CHANGELOG file".

.github/workflows/Release.yml
```diff
@@ -0,0 +1,87 @@
...
+      - name: Making relocatble CMake Install
+        run: |
+          cmake --install build/
+
+      - name: Zipping install directory (Linux/Mac)
+        if: runner.os != 'Windows'
+        working-directory: build/install
+        run: zip -r ../../ponca-${{ matrix.config.name }}.zip .
+
+      - name: Zipping install directory (Linux/Mac)
```
nmellado:
```suggestion
      - name: Zipping install directory (MSVC)
```

nmellado (approved):

Issue #106, opened 2023-08-03
Implement Basket Auto Diff


nmellado:
Add class looking like

```cpp
    template < class P, class W, int Type,
            BASKET_TP(0), BASKET_TP(1), BASKET_TP(2), BASKET_TP(3), BASKET_TP(4), BASKET_TP(5), BASKET_TP(6), BASKET_TP(7), BASKET_TP(8), BASKET_TP(9), BASKET_TP(10), BASKET_TP(11) >
    class BasketAutoDiff
            : public Ext0::DDerType<P,W,Type,PrimitiveDer<P,W,Type,Ext11<P,W, Ext10<P,W, Ext9<P,W, Ext8<P,W, Ext7<P,W, Ext6<P,W, Ext5<P,W, Ext4<P,W, Ext3<P,W, Ext2<P,W, Ext1<P,W, Ext0<P,W, PrimitiveBase<P,W> > > > > > > > > > > > > > > {

    };
    /// usage: BasketAutoDiff<P, W, ScaleDer, OrientedSphereFit>;
```

In extensions, do 
```cpp
    template < class D, class W, int Di, typename TT>
    using DDerType = OrientedSphereDerImpl<D,W,Di,TT>;
}; //class OrientedSphereFitImpl
```

nmellado:
@azaleostu if you want to play a bit with template stuff, this issue could be a good extension to your PR #85 
😇

nmellado:
@azaleostu could you push back your proposal to a new PR ?

BDoignies:
@nmellado Is this PR still relevant ? The autodiff title is a bit misleading to me. Judging by the opener, the API changed quite a bit. 

In the same theme, I guess we could make each extension "DerivativeAware". They provide a way to know what is their derivative type. In my mind: 

```c++
template<class P, class NF, class T>
struct GLSParam : public T
{
      template<int type, class _P, class _NF, class _T>
      using Derivative = GLSDer<type, _P, _NF, _T>;
};
```

And we could automatically build the BasketDiff from this information (though it will rely on some template tricks in cases this is not defined). 
The downside is that it lowers the flexibility: the user would not be able to provide a custom derivative to GLSParam. This can be solved by creating a "BasketDiffDefault".

nmellado:
> In the same theme, I guess we could make each extension "DerivativeAware". 

Yes that's the idea.

> And we could automatically build the BasketDiff from this information (though it will rely on some template tricks in cases this is not defined).

[...]

PR #334, opened 2026-05-28
Autodiff 

BDoignies:
Proposal for #106 , which adds the ability for Basket Extensions to indicate their derivative type.

For now, this first commits only introduce the Basket template side (`defines.h` will not have the type written yet). In its current form, each type would be required to provide a `DiffExtType` (cannot use `DiffType`, but this could be changed) in the following form:

```c++
template <typename _P, typename _NF, int _Type, typename _T> // P and NF could be reused from the base class
using DiffExtType = TheDerivativeClass<_P, _NF, _Type, _T>;
```

`NoDerivative` is a new type that indicates there are no derivatives. I am not exactly sure whether this type is needed or not. With concepts, we might be able to simply skip the ones for which `T::DiffExtType` does not exist.

One mandatory change however: `Base` must be accessible. The main reason is that when the extension list is provided, the types are not complete: you give a `CovariancePlaneFit`, not a `CovariancePlaneFit<Point, Filter, OTHER>`. Hence, we cannot access any subtype; we need the complete type for this.

The solution I provide here is to recurse through the `Base` chain. Because the Basket type has effectively been instantiated, each extension is complete. This does, however, require `Base` to be public.

Any remarks @nmellado?

### DIFF
--- a/Ponca/src/Fitting/Basket/basket.h
+++ b/Ponca/src/Fitting/Basket/basket.h
@@ -95,7 +95,23 @@ namespace Ponca
             using type = Ext<BSKP, BSKNF, Type, BasketType>;
         };
 
-        /*!
+        template<typename T>
+        concept HasBase = requires { typename T::Base; };
+
+        template<int Type, typename BasketExt, typename BasketType, template <class, class, int, typename> class Ext, template<class, class, int, typename> class... Exts>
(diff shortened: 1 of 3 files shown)


nmellado:
The problem will be solved by the Factory. Closing

PR #337, opened 2026-05-29
Some fixs for CI

BDoignies:
This PR proposes a few fixes for the problematic CI. 

Goals:
- [X] Fix sycl container CI that tries to push the image (which we don't want)
- [x] Fix codevoc that tries to build the instantiate multiple times

### DIFF
--- a/.github/workflows/Coverage.yml
+++ b/.github/workflows/Coverage.yml
@@ -23,7 +23,7 @@ jobs:
         with:
           path: src
       - name: cmake
-        run: mkdir build && cd build && cmake ../src  -GNinja -DCMAKE_CXX_COMPILER=g++ -DCMAKE_C_COMPILER=gcc -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=../install -DPONCA_CONFIGURE_EXAMPLES=OFF -DPONCA_CONFIGURE_DOC=OFF -DPONCA_CONFIGURE_TESTS=ON -DPONCA_COVERAGE_TESTING=ON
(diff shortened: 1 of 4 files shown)


BDoignies:
Okay, so everything works now. 

The reason Coverage fails was because of OOM error. I am not exactly sure why this is linked to instantiate library as the error still happened when compiling individual files...

The fix I provide is to make it compile in release mode which solves the issue. It is also much faster...

While experimenting, I did benchmark the compilation of the instantiate library. It uses up to 6G of RAM... ... ... This is self explanatory I guess...

nmellado (dismissed):

nmellado:
> While experimenting, I did benchmark the compilation of the instantiate library. It uses up to 6G of RAM... ... ... This is self explanatory I guess...

One option would be to split the instantiate file into 2-3 cpp files (e.g. by scalar type ?), I guess that could reduce the RAM usage if compiled sequentially (still allowing multithreading on systems with larger RAM capacities). What do you think ? (out of the scope of this PR btw)

nmellado:
Hum, still failing...

BDoignies:
I guess the solution for now is to disable instantiation for coverage testing. I think the fact that coverage testing disable optimization (-O0 passed as a flag) is the cause problem with the huge library. Compiler might not strip enough symbols or there are too much loaded at the same time. I am not so sure how to fix this but I can try as part as the mentioned PR. Coverage is not good enough, but it runs. 

BTW for fit_monge_patch: here is a seed that fail s1780312113 on Ubuntu. The run outputs the two following files (renamed as .txt because github does not allow .xyz): 

[input.xyz.txt](https://github.com/user-attachments/files/28463325/input.xyz.txt)
[projected.xyz.txt](https://github.com/user-attachments/files/28463331/projected.xyz.txt)

The reported error is 1.09802 when the epsilon threshold is 0.893718. 

I do not know if this is linked or not, but the code does not compute the RMSE a stated in the comments but rather some kind of MAE. (The code computes sum of norms instead of root of sum of norm squared).

nmellado:
Thanks. 
2 questions:
 - do you understand the coverage drop ? Some tests have line uncovered, I don't see why.
 - do you have an idea why the coverage recompiles the lib file like this ?

BDoignies:
@nmellado 

* I fixed the instantiate.o file being compile multiple times. This was due to sources of Ponca::Instantiate being set as public. I guess the generator expected this to have some impact. I fixed this (in this PR) and made the sources private (as they should)
* As far as coverage is concerned. I have no idea. The report makes no sense to me. For instance, how can the lines be partially covered / not covered? The Fit constructor is not covered but the for loop is?  Still, I have maybe one change in mind that could be responsible for this. 
image

Some other things are just because with concepts we added some operator that are not used by tests but rather checked by concepts.

PR #339, opened 2026-06-02
[fitting] Add class to check for multipasses

BDoignies:
@nmellado As we talked about here is a class in order to ease multipass computation. 

In the end, MongePatch is the only class where this is usefull. Most classes that perform a "check" just requires to know if the result is "STABLE" or not. 

I also have a few question regarding the other FIT_RESULT values. It feels like UNSTABLE should also allow result to pass through. I just copied what was in MongePatch for now.


nmellado:
The proposal sounds good, but you have a copy/paste or merge issue in the definition of the MultiPass class.

BDoignies:
No idea what you're talking about :see_no_evil:

nmellado:
We have a test that compare single-pass covariance with a standard two-step process:
https://github.com/poncateam/ponca/blob/0702da11569631befb2f8c56996449384fc9ebcf/tests/src/fit_cov.cpp#L24-L48

Please update with your proposal.

PR #348, opened 2026-09-15
[fitting] Add hrbf fitting

nmellado:


Ponca/src/Fitting/Filters/weightKernel.h
```diff
@@ -333,4 +333,36 @@ namespace Ponca
         static constexpr bool isDDValid = true;
     }; // class GaussianWeightKernel
 
+    /*!
```
BDoignies, line 336: Given the description "this is not a weight kernel", this class should probably be in another file.

Ponca/src/Fitting/Fit/hrbf.h
```diff
@@ -0,0 +1,323 @@
...
+
+#include <Eigen/Dense>
+
+#define HRBF_REQUIREMENTS ProvidesNormal<P>
+
+namespace Ponca
+{
+
+    /*!
+     * \brief Corrected Normal Current Fit type.
```
BDoignies, line 29: Doc is copy pasted from CNC

Ponca/src/Fitting/Fit/hrbf.h
```diff
@@ -0,0 +1,323 @@
...
+                    Scalar squared_l = diff.squaredNorm();
+
+                    grad += alpha_dphi * diffNormalized;
+                    grad += bDotd_l * (ddphi * diffNormalized - diff * dphi / squared_l) + beta * dphi / l;
+                }
+            }
+            return grad;
+        }
+
+        //
```
BDoignies, line 277: Should be removed (copy-pasted from cnc)

Ponca/src/Fitting/Fit/hrbf.h
```diff
@@ -0,0 +1,323 @@
...
+            return m_eCurrentState = STABLE;
+        }
+
+        /*!
+         * \brief Compute function that iterates over a subset of sampled points from an STL-Like container.
+         * \tparam IndexRange An STL-like container storing the indices of the neighbors
+         * \tparam PointContainer An STL-like container storing the points
+         */
+        template <typename IndexRange, typename PointContainer>
+        PONCA_MULTIARCH inline FIT_RESULT computeWithIds(const IndexRange& ids, const PointContainer& points)
```
BDoignies, line 153: Now that ponca supports C++20, I suggest instead something like:

```c++
auto view = ids | std::views::transform([&](size_t i) { return points[i]; });
compute(view.begin(), view.end());
```

Performance wise this should be about the same. Transforms and pipes inherit the iterator category.

BDoignies: Now that I wrote the comment, I wonder if this could be the default for ComputeObject::compute...

Ponca/src/Fitting/Filters/weightKernel.h
nmellado: Yeah, I hesitated a lot. 
I wrote `This is not a decreasing weight function`, but it is actually a weight kernel in the end. But it is used to weight basis functions according to distance, not neighbors.

nmellado: I kept it here as has the same semantics: function and its derivatives.

Ponca/src/Fitting/Fit/hrbf.h
nmellado: yes, WIP. Keep this open.

Ponca/src/Fitting/Fit/hrbf.h
nmellado: yes, would be better :)

Ponca/src/Fitting/Fit/hrbf.h
```diff
@@ -0,0 +1,323 @@
...
+        //! \brief Is the fitted primitive ready to use (finalize has been called and the result is stable)
+        PONCA_MULTIARCH [[nodiscard]] inline bool isStable() const { return m_eCurrentState == STABLE; }
+
+        /// writing implementations of function that are required to use HRBF fitting alongside Basket)
+        ///
+        ///
+        using NeighborFilter = DistWeightFilter<DataPoint,ConstantWeightKernel<Scalar>>;
+
+        PONCA_FITTING_APIDOC_SETWFUNC
+        PONCA_MULTIARCH inline void setNeighborFilter(const NeighborFilter& ) { ; }
```
nmellado, line 312: This kind of trick is annoying.

